PR #39, opened 2020-12-15
Feature/setting savings

itschip:


jfrader (commented):
I think you are confused on how to use the localstorage as a persistent state in react.

What you need here is to read from localstorage only the first time the component loads and then just save the new state back to the react state (recoil atoms) and at the same time replicate into local storage but not read again.

We can make some pair programming or I can guide you if you agree with what I've said.

phone/src/os/phone/hooks/state.ts
```diff
@@ -9,4 +9,8 @@ export const phoneState = {
     key: 'phoneConfig',
     default: undefined,
   }),
+  forceUpdateState: atom({
```
TasoOneAsia, line 12: Alright, lets not. As @jfrader pointed out, its better to expand the LocalStorage hook itself to initiate a recoil update rather than doing this hacky method here.

phone/src/os/phone/hooks/useLocalStorage.ts
```diff
@@ -25,16 +27,31 @@ function useLocalStorage<T>(key: string, initialValue: T) {
...
+      console.log(e);
+      // We still return initVal if error is thrown
+      return initialValue;
+    }
+  };
+
   // Trying to rep useState setter pretty close just with localStorage
-  const setVal = (value: T | ((val: T) => T)) => {
+  const setVal = (localKey, value: T | ((val: T) => T)) => {
+    console.log(localKey, value);
```
TasoOneAsia, line 45: Don't have console.log's integrated into a master merge, it gets too messy. I *eventually* want to work on adding a logging class where we can more easily capture logs at different verbosity levels.

TasoOneAsia (commented):
I agree with @jfrader previous comments about a different implementation. As it stands right now this is a little too hacky and adds a good amount of tech debt that we eventually will need to fix anyways. We can discuss better implementations whenever we all get a chance.

phone/src/os/phone/hooks/useLocalStorage.ts
TasoOneAsia: If need be add a console.log that is commented but I would prefer not to.

phone/src/apps/settings/components/SettingsApp.tsx
```diff
@@ -35,34 +35,32 @@ const SubHeaderComp = (props: { text: string }) => (
...
-      setVal('zoom', val)
-    )
+    MapStringOptions(zoom, (val: string) => setZoom(val))
   );
   // Doesn't actually do anything for the time being
   const ringtones = config.ringtones.map(
-    MapStringOptions(getStorageItem('ringtone'), (val: string) =>
-      setVal('ringtone', val)
-    )
+    MapStringOptions(ringtone, (val: string) => setRingtone(val))
```
jfrader, line 63: All of these can be replaced by MapStringOptions(x, setX)

phone/src/apps/settings/components/SettingsApp.tsx
```diff
@@ -71,9 +69,7 @@ export const SettingsApp = () => {
 
   // TODO: These new settings all work
 
-  useEffect(() => {
-    // this will be used to trigger a alert that gives feedback when the a setting is changed
-  }, [setVal]);
+  useEffect(() => {}, [theme]);
```
jfrader, line 72: Delete if not used

jfrader (approved):
I would rename getStorage to useStorage as the hook naming conventions.
Also why is just theme under settings but not all the others? 
Shouldn't all of them be there?

TasoOneAsia:
This isn't a hook. In the sense that it isn't operating as a hook should. Lemme refactor and show you.

PR #53, opened 2021-01-11
Added European Portuguese as a language

HypeLevels:
This time it has no lua and it got added to i18n.ts


TasoOneAsia:
lgtm merging! Thanks Hype

PR #57, opened 2021-01-17
Lerna integration with husky for precommit checks

TasoOneAsia:
This adds husky back with the correct configuration for a monorepo setup.

The monorepo is being handled using Lerna in the root directory. Both `phone/` and `resources/` have a script called `precommit` that will get called on each commit. This script runs test suites and `lint-staged` for all staged files.

tldr: Prettier is now handled locally rather than through the CI setup.


itschip (approved):
LGTM 👍

PR #62, opened 2021-02-03
Feature/show app name on hover

erik-sn:
Enables tooltips on app names and adds some TS interfaces around app configurations.

### DIFF
--- a/phone/src/ui/components/GridMenu.tsx
+++ b/phone/src/ui/components/GridMenu.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
 import { AppIcon } from './AppIcon';
-import { Grid, makeStyles } from '@material-ui/core';
+import { Grid, makeStyles, Tooltip } from '@material-ui/core';
 import { Link } from 'react-router-dom';
 import { useKeyboard } from '../../os/keyboard/hooks/useKeyboard';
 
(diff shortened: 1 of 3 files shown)


jfrader (approved):
Nice work!

itschip:
LGTM 👍

itschip (approved):

TasoOneAsia:
Wait a second does this pass when the DOM is resized @erik-sn?

I had previously implemented this but when zoom property is changed the portal coordinates don't recalculate properly

phone/src/ui/components/GridMenu.tsx
```diff
@@ -35,25 +43,23 @@ export const GridMenu = ({ items, Component = AppIcon }) => {
...
-            //   title={item.id}
-            //   placement='top'
-            //   arrow
-            //   classes={{ tooltip: classes.tooltip }}
-            //   PopperProps={{
-            //     anchorEl={}
-            //   }}
-            // >
-            <Grid
+            <Tooltip
```
TasoOneAsia, line 46: Portals are broken when using the CSS zoom property, as React's underlying render is not updated to match the dimensions of the DOM element in which the property `zoom` is applied to. See the following image for an example:

Dom Layout

TasoOneAsia:
As a result of these breaking changes, we should probably revert the commit, what do you guys think? @erik-sn @itschip @jfrader

RockySouthpaw:
I tested this in game and can confirm with taso that unless the zoom is set to 100%, this doesn't work as it doesn't scale with various zoom levels.

erik-sn:
Yep can be reverted, I didn't know about the zoom issue. I will take a more in depth look at that.

PR #63, opened 2021-02-04
Organized wallpapers, and removed non existing ones.

RockySouthpaw:
Pretty sure the weeb.png and weeb,gif were from chip's server and he committed them. I also removed the 3 wallpapers that were removed from the default.json previously.


XiosBombay:
LGTM

PR #64, opened 2021-02-05
fix twitter for light theme

erik-sn:
Removes coded background and text colors and falls back to theme. Minor styling updates.

image

### DIFF
--- a/phone/src/apps/twitter/components/tweet/TweetMessage.tsx
+++ b/phone/src/apps/twitter/components/tweet/TweetMessage.tsx
@@ -5,7 +5,7 @@ import { TextField } from '@material-ui/core';
 import { usePhone } from '../../../../os/phone/hooks/usePhone';
 import { getNewLineCount } from '../../utils/message';
 
-const useStyles = makeStyles(() => ({
+const useStyles = makeStyles({
   textField: {
     flex: '1 1 100%',
     padding: '10px 15px',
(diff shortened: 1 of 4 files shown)


phone/src/ui/components/Modal.tsx
```diff
@@ -5,7 +5,6 @@ import { Paper } from '@material-ui/core';
 const useStyles = makeStyles(() => ({
```
TasoOneAsia, line 5: Should be able to just pass an object without using an anon function when the `ThemeObject` does not need to be passed.
```ts
makeStyles({})
```

phone/src/apps/twitter/components/tweet/TweetMessage.tsx
```diff
@@ -4,17 +4,15 @@ import { makeStyles } from '@material-ui/core/styles';
 import { TextField } from '@material-ui/core';
 import { usePhone } from '../../../../os/phone/hooks/usePhone';
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles(() => ({
```
TasoOneAsia, line 7: Same as above
```ts
makeStyles({})
```

TasoOneAsia (changes requested):
Aside from small changes for cleaning up `makeStyles` when not passing `ThemeObject`, LGTM 👍

RockySouthpaw:
Looks fine to me as well. 

https://i.imgur.com/cbCcSch.png

RockySouthpaw (approved):

erik-sn:
@TasoOneAsia good call, updated!

PR #65, opened 2021-02-05
Bug/twitter char limit

erik-sn:
- updates character limit to 160
- adds a maximum new line limit (set to 10) so that not all 160 characters can be `\n` to consume a crazy amount of space
- Sets a max height on the new tweet TextField with a scrollbar so large messages don't overrun the screen

FiveM_GTAProcess_CIzZTc0Gm7

### DIFF
--- a/phone/src/apps/twitter/components/tweet/TweetMessage.tsx
+++ b/phone/src/apps/twitter/components/tweet/TweetMessage.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from '@material-ui/core/styles';
 import { TextField } from '@material-ui/core';
 import { usePhone } from '../../../../os/phone/hooks/usePhone';
+import { getNewLineCount } from '../../utils/message';
 
 const useStyles = makeStyles((theme) => ({
   textField: {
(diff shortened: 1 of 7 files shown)


RockySouthpaw:
Will test shortly

phone/src/apps/twitter/components/tweet/TweetMessage.tsx
```diff
@@ -1,14 +1,17 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from '@material-ui/core/styles';
 import { TextField } from '@material-ui/core';
 import { usePhone } from '../../../../os/phone/hooks/usePhone';
+import { getNewLineCount } from '../../utils/message';
 
 const useStyles = makeStyles((theme) => ({
```
jfrader, line 8: U can remove theme here and just pass the style object like so:
```
makeStyles({ root: { ... } });
```

jfrader (approved):
LGTM

RockySouthpaw (approved):
Works as intended and no issues arise from this. Theres an interesting feature where if you do something like:
> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

it will only send it as 1 line instead of populating the lines below it. 

https://i.imgur.com/5tKUQtf.png

itschip:
@erik-sn Look at the last comment from Rocky about 'overflow' when you do a one-liner of long text. :)

PR #70, opened 2021-02-06
Feat/inject debug - Use new abstraction to clean up debug data

TasoOneAsia:


phone/src/apps/twitter/components/TwitterApp.tsx
```diff
@@ -35,6 +36,17 @@ const useStyles = makeStyles(() => ({
 
 const TWEETS_REFRESH_RATE = 15000; // TODO move this to twitter config
 
+InjectDebugData([
```
itschip, line 39: We don't really need this, but whatever

itschip (approved):
LGTM 👍

phone/src/apps/twitter/components/TwitterApp.tsx
TasoOneAsia: Include or not, I moved all the debug events from Phone.tsx

jfrader (approved):

PR #72, opened 2021-02-07
Improved install instructions and cleanup.

RockySouthpaw:


jfrader (commented):
Why powershell? It's not necessary

RockySouthpaw:
> Why powershell? It's not necessary

This was a remnant of the previous guide, I only added a period. It should still be mentioned though as just the word “terminal” would be vague. Its just assumed prerequisites to make the guide process easier. It’s better to troubleshoot/walk people through it if they’re using the same terminal program anyways. 

The actual build instructions included visual studio’s terminal as an option 
> 1. Open **Windows Powershell** or a terminal in **Visual Studio Code** 
so I’ll look into expanding it some more I guess or state that perhaps it’s not necessary, but the guide assumes you’re using it or the terminal in Vidual studio.

 For the sake of future support, it would be best to mention a specific one and if they’re a dev they would obviously know they can use whatever.

PR #73, opened 2021-02-07
Remove client phoneConfig event

jfrader:


TasoOneAsia:
Conflicting

jfrader:
Fixed conflicts @TasoOneAsia

phone/src/apps/twitter/components/TwitterApp.tsx
```diff
@@ -152,3 +152,17 @@ export const TwitterApp = () => {
...
+    method: 'getOrCreateTwitterProfile',
+    data: {
+      id: 1,
+      profile_name: 'Elon Musk',
+      identifier: 'elonmusk',
+      createdAt: Date.now() / 1000,
+      updatedAt: Date.now() / 1000,
+    },
+  },
+]);
```
TasoOneAsia, line 168:
```suggestion
```

TasoOneAsia, line 168: Profile already injected at top

```suggestion
```

phone/src/common/typings/config.ts
```diff
@@ -25,7 +25,7 @@ interface BankConfig {
   showNotifications: boolean;
 }
 
-export interface ServerConfig {
+export interface IServerConfig {
```
TasoOneAsia, line 28: Whats our standard for the naming on these interfaces? Are we in the `I` prefix camp or against it, because I see inconsistent naming across the phone.

phone/src/os/phone/hooks/usePhone.ts
```diff
@@ -1,14 +1,14 @@
...
   visibility: boolean;
-  config?: ServerConfig;
+  config?: IServerConfig;
 }
 
 export const usePhone = (): IUsePhone => {
   const visibility = useRecoilValue(phoneState.visibility);
-  const config = useRecoilValue(phoneState.phoneConfig);
-  return { visibility, config };
+  return { visibility, config: ServerConfig as any };
```
TasoOneAsia, line 13: This will assign `config` a type of IServerConfig rather than any as previous

```suggestion
  const config = ServerConfig as IServerConfig
  return { visibility, config };
```

TasoOneAsia, line 13: This will assign `config` a type of IServerConfig rather than an any

```suggestion
  const config = ServerConfig as IServerConfig
  return { visibility, config };
```

TasoOneAsia (changes requested):

phone/src/os/phone/hooks/usePhone.ts
jfrader: This has issues with enum and not string types because it's coming from json. That's why I left any first.

phone/src/common/typings/config.ts
jfrader: I use I for interfaces or types that are actual interfaces. and E for encima for example.

phone/src/os/phone/hooks/usePhone.ts
TasoOneAsia: Ah I see the use case then, if we are going to use Enum types should we just move it a .ts file?

phone/src/common/typings/config.ts
TasoOneAsia: From now on I'll start using the I / E prefix as I hadn't been before

phone/src/os/phone/hooks/usePhone.ts
jfrader: Let's merge and improve that in another iteration.

TasoOneAsia (approved):

PR #74, opened 2021-02-07
Fix twitter focusing bug

erik-sn:
[Board Link](https://app.gitkraken.com/glo/view/card/230b3e33ec9a4c27810236b0c45e7073)

- Fixes a bug where opening media link caused that field to permanently focus
- Adds some render optimizations to the fields to prevent excessive re-renders (still other opportunities to resolve as well)


RockySouthpaw:
Will test the twitter portion in a few minutes

RockySouthpaw (commented):
Tested within FiveM and the bug is resolved. 👍🏻

TasoOneAsia (approved):

PR #80, opened 2021-02-09
Refactored Contacts to use Routes and not a Modal

TasoOneAsia:
This change removes the modal and adds a route of `contacts/:ContactId`, where ContactID is the contact id
for the stored contact. It also opens up the possibility for other apps to directly route to a contact easily. This can be expanded with parameters at a later point to allow for those apps to also pass specific inputs autofilled for the user.


jfrader (approved):
Nice work!

itschip (approved):
LGTM 👍 I just wish we could do the styling in separate files, but I guess it doesn't make that big of a difference anyways.

TasoOneAsia:
> styling in separate files,

Was going to do this but judging on how everyone seems to be doing styling in this way,  we probably want to be consistent with styling being next to the component just for maintainability sake's. We really should standardized this tbh.

PR #85, opened 2021-02-12
Translation: Update fr.json

Oldarorn:
New update !


jfrader (approved):

RockySouthpaw:
In the future state what you added in the description :p. At first glance I was wondering what the new update was but I realized it was translation. 

Example:
- Added translation Twitter, sellout and other app names.
- Added translation for notes.
- Added date translation.

PR #92, opened 2021-02-15
Feature/message from contacts

jfrader:


TasoOneAsia:
@jfrader Does the number actually need to be valid? Testing in game with a non-valid number breaks this.

jfrader:
Ok I'll handle that case tomorrow, thanks!

itschip:
@jfrader Do you think it would be a _good_ idea to check if there is already an existing conversation, whereas we then route the user to said conversation, instead of always routing to '/new'?

PR #93, opened 2021-02-15
refactor(server/source-handling): Altered how we handle source

TasoOneAsia:
These changes are made in case source is not guaranteed to remain the same

Due to how JS should operate, this shouldn't be an issue but this just ensures that


itschip (approved):
LGTM 👍 Liked what you did with the 'getIdentifier'!

PR #95, opened 2021-02-15
feat(client/net): Promisfied net events

TasoOneAsia:
This new Feature allows for us to wrap net events in a promise. Here are some examples.

## Client
```ts
try {
  const result = await ClUtils.emitNetPromise(
    'npwd:promiseEvent',
    'arg1',
    'arg2'
  );
  console.log('Resolved:', result);
// If the promise is rejected, we catch it here
} catch (e) {
  console.error('Promise Rejected:', e);
}
```
## Server
This is how we listen to these promised events on the server. The first param to the callback will always be the name of the event that we must emit back to the client. The rest of the params, are whatever was passed to `emitNetPromise`. When responding, we must use the passed eventName, the source, and either the result or an error. The following example is a successful server transaction, where the data returned is 'Amazing data returned'.
```ts
onNet('npwd:promiseEvent', (eventName: string, arg1: string, arg2: string) => {
  const _source = (global as any).source;

  emitNet(eventName, source, 'Amazing data returned');
});
````

If we wanted to inform the client an error has occurred, and wish to reject the promise, we do this:
```ts
onNet('npwd:promiseEvent', (eventName: string, test1: string, test2: string) => {
  const _source = (global as any).source;

  emitNet(eventName, source, null, 'Error has occurred, you suck');
});
```
The first argument passed to the return event will always be data that was successfully retrieved according to the event's scope. The second argument is an error to be sent to the client, this will be passed to the Promise rejection.

### DIFF
--- a/resources/client/cl_utils.ts
+++ b/resources/client/cl_utils.ts
@@ -1,0 +1,59 @@
+import { uuidv4 } from '../utils/fivem';
+
+interface ISettings {
+  promiseTimeout: number;
+}
+
+interface ISettingsParams {
+  promiseTimeout?: number;
+}
+
+export default class ClientUtils {
+  private _settings: ISettings;
+  private _defaultSettings: ISettings = {
+    promiseTimeout: 5000,
+  };
+
+  constructor(settings?: ISettingsParams) {
+    this.setSettings(settings);
+  }
+
+  public setSettings(settings: ISettingsParams) {
+    this._settings = {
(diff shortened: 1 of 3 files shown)


itschip:
Why not use the uuid package we already have?

TasoOneAsia:
@itschip There is a relevant comment pertaining to that

itschip:
@TasoOneAsia Alright!

PR #99, opened 2021-02-16
Fix SQL import file collation

jfrader:


RockySouthpaw:
Does the other portions of the SQL have any issue? Or is it strictly twitter portion?

TasoOneAsia:
Would we be able to implicitly collate everywhere and remove the explicit collation definitions also present in other queries?

jfrader:
It was strictly to twitter.
Also, Taso, we should remove every collation and charset there, but we can improve that later, utf8 is supported everywhere
@TasoOneAsia @RockySouthpaw

TasoOneAsia (approved):

PR #100, opened 2021-02-17
Fix: Call history is now sorted.

RockySouthpaw:
Call history Is now ordered by ID so recent calls show up first instead of last. Previously it was just at the bottom.

(Courtesy of Chip)


TasoOneAsia (approved):

itschip (approved):
LGTM 👍

PR #102, opened 2021-02-17
Feature / Add working notifications with only twitter for now

jfrader:


RockySouthpaw:
Only twitter? Denied. /s

itschip:
Could you bring the phone down to only show the last notification?

jfrader:
> Could you bring the phone down to only show the last notification?

Let's improve that over next iterations, because for now, notifications don't have a fixed height.

PR #104, opened 2021-02-18
Feat/Call animations and fixed reject call being null.

RockySouthpaw:
### Feat: Call Animations
**All** animations were moved from `cl_main.ts` to `functions.ts`. Calling animations function as followed:

- If a player calls and the target **does not** answer, then no **additional** animation is played. This is done so in the event someone isn't in the server or potentially doesn't have a phone, the animation doesn't play. They continue to play the standard animation for when your phone is open. Looks better too as they're just looking at their phone to see if someone answers.

- If a player calls and the target **does** answer then BOTH will play the text to call animation. 

- When **ending** an **active** call, both parties play the call to text animation.

I adjusted the speed of the animations so they should be more fluid than gcphone was. I still think it would be best to stop the call animation and close the phone, but I didn't include that change. Calls in vehicles also have their own animations as well.

[Video for reference](https://streamable.com/rr7m1l) (Thanks @XiosBombay)

### Fixed Reject call being null
Solves this [error](https://i.imgur.com/Q7zmw39.png) thanks to @jfrader for assisting. Was as simple as changing `emitNet(events.PHONE_CALL_REJECTED, data.transmitterNumber, end);` to `emitNet(events.PHONE_CALL_REJECTED, data.phoneNumber, end);`.


XiosBombay:
LGTM

Merge

TasoOneAsia:
I feel obliged to close this PR on that video alone 😞 , but let me review it here in a little bit

resources/client/functions.ts
```diff
@@ -0,0 +1,213 @@
+import { Delay } from '../utils/fivem';
+
+let prop = 0;
+let isPhoneOpen = false;
+let propCreated = false;
+let phoneModel = 'prop_amb_phone'; // Refered to in newphoneProp function. Requires custom phone being streamed.
```
TasoOneAsia, line 6: Not sure why this was a let in the first place but lets change it while we are here

```suggestion
const phoneModel = 'prop_amb_phone'
```

resources/server/sv_call.ts
```diff
@@ -176,14 +178,18 @@ onNet(
       const currentCall = calls.get(transmitterNumber);
 
       const endTime = timestamp / 1000;
-
       await updateCall(currentCall, false, endTime);
-
+      const accepted = calls.get(pSource);
+      console.log(accepted)
```
jfrader, line 183: Remove

jfrader:
ACK

resources/client/cl_call.ts
```diff
@@ -1,10 +1,11 @@
 import events from '../utils/events';
 import { ICall } from '../../phone/src/common/typings/call';
+import {phoneCallStartAnim, phoneCallEndAnim} from './functions';
 
 const exp = (global as any).exports;
 
-RegisterNuiCallbackType(events.PHONE_INITIALIZE_CALL);
-on(`__cfx_nui:${events.PHONE_INITIALIZE_CALL}`, (data: any, cb: Function) => {
+RegisterNuiCallbackType(events.PHONE_INITIALIZE_CALL); // Fires when the call is started.
+on(`__cfx_nui:${events.PHONE_INITIALIZE_CALL}`, async (data: any, cb: Function) => {
```
TasoOneAsia, line 8: This shouldn't need to be async
```suggestion
on(`__cfx_nui:${events.PHONE_INITIALIZE_CALL}`, (data: any, cb: Function) => {
```

jfrader (approved):
ACK

resources/server/sv_call.ts
TasoOneAsia: If you want to do debug logging using the logger instance
```suggestion
      callLogger.debug(`Call for ${pSource} is accepted`)
```

or something like that, otherwise no `console.log()` on the server

resources/client/functions.ts
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * * * * * * * * * * * * */
+
+const newPhoneProp = async () => {
+  removePhoneProp(); //deletes the already existing prop before creating another.
+  if (!propCreated) {
+    RequestModel(phoneModel);
+
+    while (!HasModelLoaded(phoneModel)) {
+      await Delay(1);
```
TasoOneAsia, line 20: Honestly might be a good idea to abstract this like we did with `loadDict` but its fine for now.

resources/client/cl_call.ts
```diff
@@ -89,7 +93,6 @@ on(`__cfx_nui:${events.PHONE_END_CALL}`, (data: any, cb: Function) => {
 
 onNet(events.PHONE_CALL_WAS_ENDED, () => {
```
TasoOneAsia, line 94: I may have missed the conversation revolving around this but whats the reason for not just attachingthe anim here, instead of adding another net event.

TasoOneAsia (changes requested):
Overall lgtm from first glance codewise, what this PR is telling me tho is that the calling flow needs to be seriously looked at. I am going to spend some time thinking about the best way to sync our state.

TasoOneAsia:
Also whoever does eventually merge this ***pretty please***, squash and merge with `feat(client/anim): Added animation flow for calls`

resources/client/cl_call.ts
RockySouthpaw: It would trigger for the person who's being called regardless if they answered or not if done this way. The workaround was to create the net event and trigger it server side if the call was active.

resources/client/functions.ts
RockySouthpaw: Will make note of this to do later then

resources/server/sv_call.ts
RockySouthpaw: I already removed the log, but can add it when the calling flow is re done.

PR #105, opened 2021-02-18
Enhance/settings

TasoOneAsia:


RockySouthpaw:
Looks great to me. Tested all the functionality and the copying of your number works as well.

https://i.imgur.com/KgzE61d.png

This does make me realize Tasp-dark needs a slight additional variance between the 2 colors as it's a bit hard to see where the settings and nav bar ends.

jfrader:
LGTM

jfrader:
It would be awesome if we could merge this config with the one at root directory and make them .TS and type it.
What do you think @itschip @RockySouthpaw 

This way we would only have 1 config file for everything and as it is .ts we can strictly type it for better usage.

At least i would merge both into a json if not ts.

RockySouthpaw:
For some reason, the blue phone case doesn't work in game as it shows up blank. I've tried this branch a total of 3 times now and the issue occurs each time. Not sure the cause.

itschip (commented):
Works just fine, but I think we should set some default values?

TasoOneAsia:
Will refactor onto latest to avoid all these conflicts in Phone.tsx

PR #109, opened 2021-02-21
Feature / Notes and Dialer theme providers

jfrader:


RockySouthpaw:
Testing now

RockySouthpaw:
Message overflow looks a lot better./ 👍🏻 Now it looks as such:
image

Couple of things:
- Delete icon looks more pink then red on the notes. Should keep uniformity between all delete icons i feel like as contacts looks like such:
image
- Revert the 2 app logo color change as it clashes with the others.  (Already discussed but adding it here cause reasons kek.)
image

Other then those 2, LG2M

RockySouthpaw:
## Missing Dialer Features
- A way to delete an entered value. It would need to be a UI element in my opinion since everything else is that way. 
- Dialer doesn't auto populate a dash based off the configured numbers. (Default is a dash like XXX-XXXX)

## Feedback
I personally don't see a reason to add a contact button on the dialer as the contacts page is 1 click over but that's just *my opinion*. Regardless, it doesn't hurt anything to add it. 🤷🏻 Would be neat if it took what was in the text field and auto populated it to the entry tbh. It should also be added to the contacts tab as well since you can't create a contact there at all.

I also think the dialer could be approached differently. I see no reason to include the dash as a button once it is auto generated. It really takes away design options as the rows are no longer even. I feel like anyways.

I've created some mock ups to help demonstrate what I'm talking about:
Dialer Design 1

dialer 2

I personally prefer the second one without a dash. You could also keep your current design and replace the dash button with calling, replace calling button with backspace and keep add contact how it is.

PR #114, opened 2021-02-23
Fix/number generating

RockySouthpaw:
Fixed number not generating on player joined and adjusted the SQL. I also removed the instances of _identifier = identifier because it wasn’t needed. (Kidz helped a lot with this <3.)

Squash please.
[Phone number doesn't auto generate a number.](https://app.gitkraken.com/glo/card/0ae48fae03e54b6d901cddf36a13b619)

### DIFF
--- a/resources/import.sql
+++ b/resources/import.sql
@@ -1,6 +1,6 @@
 -- comment this out if your server has a different table/location for phone_number
 -- or if it already exists in the users table
-ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) NULL DEFAULT NULL;
+ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) DEFAULT NULL;
 INSERT INTO `items` (`name`, `label`, `weight`, `rare`, `can_remove`) VALUES ('phone', 'Phone', 1, 0, 1);
 INSERT INTO `items` (`name`, `label`, `weight`, `rare`, `can_remove`) VALUES ('dphone', 'Destroyed Phone', 1, 0, 1);
 
--- a/resources/server/sv_main.ts
+++ b/resources/server/sv_main.ts
@@ -35,20 +35,27 @@ function getRandomPhoneNumber() {
     // The numbers inside {} in replace() can be changed to how many digits you want on each side of the dash.
     // Example: 123-4567
   }
-
   mainLogger.verbose(`Getting random number: ${randomNumber}`);
 
   return randomNumber;
 }
 
 async function generatePhoneNumber(identifier: string) {
-  let phoneNumber = await usePhoneNumber(identifier);
-
-  if (!phoneNumber) {
-    phoneNumber = getRandomPhoneNumber();
+  if (!await usePhoneNumber(identifier)) {
+    let existingId;
+    let newNumber;
+    do {
+        newNumber = getRandomPhoneNumber();
+        try {
+          existingId = await getIdentifierByPhoneNumber(newNumber)
+        } catch(e) {
+          existingId = false
+        }
+    } while(existingId);
+    mainLogger.verbose(`Inserting number into Database: ${newNumber}`);
     const query = 'UPDATE users SET phone_number = ? WHERE identifier = ?';
-    await pool.query(query, [phoneNumber, identifier]);
-  } 
+    await pool.query(query, [newNumber, identifier]);
+  }
 }
 
 async function getCredentials(identifier: string): Promise<string> {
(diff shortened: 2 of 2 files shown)


jfrader:
LGTM didn't Test

jfrader (approved):
LGTM, didn't test.

itschip:
This is not good enough. We're not checking if the phone number already exists among any other users.

RockySouthpaw:
You want people to not have the same number? What is this an RP phone? Gosh chip...

Ye I’ll figure it out and add it. Thank you for informing me <3

resources/import.sql
```diff
@@ -1,6 +1,6 @@
 -- comment this out if your server has a different table/location for phone_number
 -- or if it already exists in the users table
-ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) NULL DEFAULT NULL;
+ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) DEFAULT NULL;
```
erik-sn, line 3: Can we add something like `IF NOT EXISTS` as we do in the table generation so that people don't have to edit this file if they already have this column?

resources/server/sv_main.ts
```diff
@@ -42,19 +42,13 @@ function getRandomPhoneNumber() {
...
-    if (!phoneNumber) {
-      phoneNumber = await getRandomPhoneNumber();
-      const query = 'UPDATE users SET phone_number = ? WHERE identifier = ?';
-      await pool.query(query, [phoneNumber, _identifier]);
-    } else {
-      break;
-    }
-  } while (id);
+  if (!phoneNumber) {
+    phoneNumber = getRandomPhoneNumber();
```
erik-sn, line 48: We should check that this number doesn't already exist

erik-sn (changes requested):
Possible edge case:

A server already exists and everyone already has phone numbers they don't want to change, but they're not in `user.phone_number`. The server owner could perform the migration themselves but  is there a way to handle it here?

resources/import.sql
RockySouthpaw: So as far as I understand from researching this, different mysql versions handle this whole `IF NOT EXIST` differently so there wouldn't be a universal case for this. We can always include a way to drop the SQL i guess:
`ALTER TABLE `users` DROP COLUMN `phone_number`;`. 

I agree there needs to be a better way to do this since in the past with gcphone just the one column not being proper caused issues if they came from another phone.

resources/server/sv_main.ts
RockySouthpaw: Yea he did :p We all had discussed this in discord on the best way to implement it as the way it was before, for whatever reason, didnt work for everyone. I don't remember what the conclusion was though and i havent had time to implement it.

RockySouthpaw: Fixed

RockySouthpaw:
@itschip Please review again.

resources/server/sv_main.ts
itschip: @RockySouthpaw Did you add this back again? Or is this something we'll fix later? As this could be fixed in the PR, no?

jfrader:
> Possible edge case:
> 
> A server already exists and everyone already has phone numbers they don't want to change, but they're not in `user.phone_number`. The server owner could perform the migration themselves but  is there a way to handle it here?

I agree with this. We should be able to migrate this at some point

jfrader:
> @itschip Please review again.

I think u didn't push latest changes

RockySouthpaw:
HAHAHA I DIDNT PUSH

resources/server/sv_main.ts
```diff
@@ -65,17 +66,10 @@ async function getCredentials(identifier: string): Promise<string> {
...
-  } catch(e) {
-    mainLogger.error(`Failed to generate a phone number, ${e.message}`, {
-      source: _source,
-    })
-  }
-})
+onNet('esx:playerLoaded', async (playerId: number, xPlayer: any) => {
+  const identifier = xPlayer.identifier;
+  await generatePhoneNumber(identifier);
+});
```
itschip, line 72: Might want to add a try catch with a `mainLogger.error` to provide a error log.

RockySouthpaw: We have a error log in `generatePhoneNumber` already so wouldn’t this be redundant? Just for phone number creation we have a catch there and on `generateRandomNumber` 🤔

itschip (approved):
Tested and worked! LGTM 👍

PR #117, opened 2021-02-25
Remove dependency comments and use useRef instead for useSound hook

jfrader:


RockySouthpaw:
I assume this just needs code review?

TasoOneAsia (approved):

PR #119, opened 2021-02-28
Add calling and notification sounds

jfrader:


RockySouthpaw (approved):
LG2M, Tested with chip

PR #120, opened 2021-02-28
Fix/ Message error pertaining to identifier 

RockySouthpaw:


itschip:
Looks good! Just merge it

PR #125, opened 2021-03-01
tweak(phone/snackbarprovider): remove redunant function

itschip:


phone/src/ui/providers/SnackbarProvider.tsx
```diff
@@ -16,10 +16,8 @@ function SnackbarProvider({ children }) {
     return () => clearTimeout(timer.current);
   }, [alert]);
 
-  const addAlert = (value) => setAlert(value);
-
   return (
-    <SnackbarContext.Provider value={{ addAlert, alert }}>
+    <SnackbarContext.Provider value={{ alert, addAlert: setAlert }}>
```
TasoOneAsia, line 20: Its probably cleaner to just use setAlert and change all occurrences rather than changing it in this manner.

itschip: fuck off

itschip: but...fine

PR #129, opened 2021-03-02
[Bug] Yarn watch isn't working

erik-sn:
"watch" in the phone package.json allows rebuilds on any code changes. This doesn't seem functional as of [this commit](https://discord.com/channels/791854454760013827/791854455468458042/816454853621710868) which excludes `dev.js` from the `fxmanifest.lua`.

This change just wildcards JS files so both `yarn build` and `yarn watch` are functional.


itschip:
LGTM 😁

PR #132, opened 2021-03-03
paste from clipboard fixed

r1OG:
This is my attempt of fixing the paste from clipboard. Modal for CTRL+V removed.
[Fix: Paste from clipboard](https://app.gitkraken.com/glo/card/09c686ffa87a497a863858e474cace9b)


phone/src/apps/messages/components/modal/MessageImageModal.tsx
```diff
@@ -57,17 +64,21 @@ export const MessageImageModal = ({ isOpen, messageGroupId, onClose, image }: IP
   }, [image, sendFromQueryParam]);
 
   const sendFromClipboard = (event) => {
-    if (event.key === 'Enter') {
-      sendImageMessage(message);
-    }
+    navigator.clipboard.readText().then((text) => {
+      if (isImage(text)) {
+        sendImageMessage(text);
```
jfrader, line 69: You could `setQueryParamImagePreview(text)` so it triggers the modal first to preview before you send, and the rest is already taken care after, because the preview modal has the sendImage action onClick.

jfrader (changes requested):
Everything looks good, just a small change to preview the image before sending.

phone/src/apps/messages/components/modal/MessageImageModal.tsx
jfrader: (BTW, `queryParamImagePreview` and `setQueryParamImagePreview` seems like a bad name now we should put something more generic like `imagePreview`, `setImagePreview`)

phone/src/apps/messages/components/modal/MessageImageModal.tsx
```diff
@@ -27,6 +29,11 @@ export const MessageImageModal = ({ isOpen, messageGroupId, onClose, image }: IP
   const [pasteVisible, setPasteVisible] = useState(false);
   const [queryParamImagePreview, setQueryParamImagePreview] = useState(null);
 
+  const isImage = (url) => {
```
jfrader, line 32: This may be better in `utils/isImage.ts` as its used in more than 2 places I think

RockySouthpaw:
Will test sometime tonight 👍🏻

TasoOneAsia:
To add on to Kidz check out #133, for our general util for image validation.

jfrader (approved):

RockySouthpaw:
FiveM_GTAProcess_qYsIUJaI8a
FiveM_GTAProcess_UjxBvCDcIk

Here's the error when pasting. Just opened messages > clicked a group > entered text into text field > copied it > deleted it > and pasted it via the prompt.

TasoOneAsia:
Unfortunately FiveM's CEF implementation doesn't grant access to the Clipboard API 🤷

RockySouthpaw:
> Unfortunately FiveM's CEF implementation doesn't grant access to the Clipboard API 🤷

So you can copy something automatically to someone's clipboard (coords resources) but can't fetch what they have copied? lol

TasoOneAsia:
> So you can copy something automatically to someone's clipboard (coords resources) but can't fetch what they have copied? lol

Yes, probably for security concerns. Pasting has always been something that needed granted perms for, even on regular browsers. We could always make a feat req to implement these perms in the same manner as `getUserMedia` is currently but not sure what FiveM would say 🤷‍♂

itschip (commented):
As it has been mentioned, we need to do something else than using said API for pasting.

itschip:
We need to go back the previous step where we had a text field to paste in the link. From there we'll have to check if its an valid image, and then open the share modal?

RockySouthpaw:
> We need to go back the previous step where we had a text field to paste in the link. From there we'll have to check if its an valid image, and then open the share modal?

Currently we auto parse images when they're sent as plain links so to be honest, we could just abstract the whole "paste from clipboard" thing. And explain that our phone auto parses URL.

TasoOneAsia:
Stale to be looked at again later

PR #133, opened 2021-03-03
tweak(phone/utils): Refactored image validation to promise from cb

TasoOneAsia:


jfrader (approved):
I like this approach but we should move it to a utils,  there's a lot of places where we validate that.

Including an open PR (clipboard)

TasoOneAsia:
@jfrader `phone/src/common/utils`?

PR #134, opened 2021-03-04
enhance(phone/settings): General enhancements

TasoOneAsia:
General enhancements to settings app

### DIFF
--- a/phone/src/Phone.tsx
+++ b/phone/src/Phone.tsx
@@ -59,21 +59,21 @@ function Phone() {
             className="Phone"
             style={{
               transformOrigin: 'right bottom',
-              transform: `scale(${settings.zoom}`,
+              transform: `scale(${settings.zoom.value}`,
               bottom,
             }}
           >
             <div
               className="PhoneFrame"
               style={{
-                backgroundImage: `url(./media/frames/${settings.frame})`,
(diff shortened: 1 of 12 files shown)


jfrader (approved):
Lgtm

TasoOneAsia:
These our ***BREAKING*** changes to setting schema, if you do not reset settings, the phone will crash until further fallbacks are added for the SoundProvider.

erik-sn (approved):
We should probably implement a similar `Option` interface for the rest of the phone configuration at some point too.

jfrader:
Yes i think everything should be in the same place in the root and get copied over or just imported.

TasoOneAsia:
> Yes i think everything should be in the same place in the root and get copied over or just imported.
> @erik-sn @TasoOneAsia

To add, I actually think it would be best to have an JSON ignored from build that can be referenced by scripts. This way, when we get to the release stage, we can pre-webpack server/client files and still allow for end users to change the config. Also forcing a rebuild when changing the config is not gonna yield good results for the majority of the FiveM community  **"Developer"** base.

RockySouthpaw:
Aside from the super small issue with notifications being highlighted in verse for twitter, it looks good to me. 👍🏻

PR #135, opened 2021-03-04
tweak(phone/snackbar-provider): Altered API

TasoOneAsia:
Change to Snackbar message property to allow for any string rather than a translation key

**New**
```jsx
export const Snackbar = () => {
  const classes = useStyles();
  const { alert } = useSnackbar();

  return (
    <div className={classes.root}>
      {alert ? <Alert severity={alert.type}>{alert.message}</Alert> : null}
    </div>
  );
};
```

### DIFF
--- a/phone/src/apps/bank/hooks/useBankService.ts
+++ b/phone/src/apps/bank/hooks/useBankService.ts
@@ -4,17 +4,26 @@ import { useTransactions } from './useTransactions';
 import { useCredentials } from './useCredentials';
 import { useSetRecoilState } from 'recoil';
 import { useBankNotification } from './useBankNotification';
-import { useSnackbar } from '../../../ui/hooks/useSnackbar';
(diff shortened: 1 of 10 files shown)


TasoOneAsia:
Could probably be cleaned up a little bit with a hook abstraction, but wanted to get this done so I could use the Snackbar provider in #134

erik-sn (approved):
I think this makes much more sense than trying to do the translation in the generic component.

jfrader (approved):

jfrader:
@TasoOneAsia I would probably use useCallback for everything that goes into useNuiEvent, i changed messages one but the other `handleAddAlert` were not changed.

TasoOneAsia:
@jfrader 🍪 as promised

jfrader:
> @jfrader 🍪 as promised

I want a real cookie now :(

PR #137, opened 2021-03-04
Tweak/marketplace additions

itschip:
- Reports to discord 
- Delete listings

Might have to wait with merging, as we are changing prefix on alerts. And...I don't know how they look like.
[Tweak: Marketplace Additions](https://app.gitkraken.com/glo/card/b221c5464d124d60ae4ceebb14bbc0cc)


resources/import.sql
```diff
@@ -70,6 +70,7 @@ AVG_ROW_LENGTH = 4096;
 CREATE TABLE IF NOT EXISTS npwd_sellout_listings (
   id int(11) NOT NULL AUTO_INCREMENT,
   identifier varchar(255) DEFAULT NULL,
+  player varchar(255) DEFAULT NULL,
```
TasoOneAsia, line 73: What is player representing here, identity name, or?

Edit: Ah, now that I see the type is a string, this makes me think it's the identity name. We might want to rename that column to something more similar to that.

2nd Edit: Now I see you are grabbing playerName from the GetPlayerName native. Might be better to rename this to username.

resources/import.sql
```diff
@@ -78,11 +79,21 @@ CREATE TABLE IF NOT EXISTS npwd_sellout_listings (
   PRIMARY KEY (id)
 )
 ENGINE = INNODB,
-AUTO_INCREMENT = 26,
+AUTO_INCREMENT = 31,
```
TasoOneAsia, line 82: Doesn't really matter but whats the reasoning for this auto_increment number 🤔

phone/src/apps/sellout/components/SelloutList/SelloutItem.tsx
```diff
@@ -74,16 +83,37 @@ export const SelloutItem = (listing) => {
...
-          <div>
-            <Button>
-              <ChatIcon className={classes.icon} />
-            </Button>
-            <Button>
-              <PhoneIcon className={classes.icon} />
-            </Button>
+          <div className={classes.actionRow}>
+            <div>
+              <Tooltip title="Send a message">
```
TasoOneAsia, line 91: Needs locale key

phone/src/apps/sellout/components/SelloutList/SelloutItem.tsx
```diff
@@ -74,16 +83,37 @@ export const SelloutItem = (listing) => {
...
+              </Tooltip>
+              <Tooltip title={listing.number}>
+                <Button>
+                  <PhoneIcon className={classes.icon} />
+                </Button>
+              </Tooltip>
+            </div>
+            <div>
+              {listing.number === myNumber ? (
+                <Tooltip title="Delete listing">
```
TasoOneAsia, line 104: Needs locale key

phone/src/apps/sellout/components/SelloutList/SelloutItem.tsx
```diff
@@ -74,16 +83,37 @@ export const SelloutItem = (listing) => {
...
+                  <Button onClick={() => handleDeleteListing(listing.id)}>
+                    <DeleteIcon />
+                  </Button>
+                </Tooltip>
+              ) : (
+                <Tooltip title="Report listing">
```
TasoOneAsia, line 110: Needs locale key

phone/src/apps/sellout/hooks/useListing.ts
```diff
@@ -1,7 +1,12 @@
...
-export const useListing = () => {
-  const listing = useRecoilValue(selloutState.listing);
-  return { listing };
+interface ListingHook {
+  listings: MarketplaceListing[];
+}
+
+export const useListing = (): ListingHook => {
+  const listings = useRecoilValue(selloutState.listing);
+  return { listings };
```
TasoOneAsia, line 11: If you want to clean up this function a little bit so you don't need an interface/object property return, pass the useRecoilValue generic a type of MarketPlaceListing[], and simply return it.
```ts
export const useListing = () => {
  return useRecoilValue<MarketPlaceListing[]>(selloutState.listing)
}
```
Up to you but I think it looks a little cleaner this way

TasoOneAsia (changes requested):

phone/src/apps/sellout/components/SelloutList/SelloutItem.tsx
```diff
@@ -1,10 +1,15 @@
 import React from 'react';
 import { makeStyles } from '@material-ui/core/styles';
-import { Button, Typography, Paper } from '@material-ui/core';
+import { Button, Typography, Paper, Tooltip } from '@material-ui/core';
 import { ListItem } from '../../../../ui/components/ListItem';
 import ChatIcon from '@material-ui/icons/Chat';
 import PhoneIcon from '@material-ui/icons/Phone';
 import { PictureResponsive } from '../../../../ui/components/PictureResponsive';
+import { MarketplaceListing } from '../../../../common/typings/marketplace';
+import DeleteIcon from '@material-ui/icons/Delete';
```
erik-niehaus, line 9: a slight pet peeve I have that I definitely should have voiced earlier in the project lol - can we put 3rd party package imports above relative/project imports?

At work someone made a slack bot that alerts a channel everytime I make a "import order" PR comment so I think I may be alone in the world with this particular bit of OCD.

resources/import.sql
```diff
@@ -78,11 +79,21 @@ CREATE TABLE IF NOT EXISTS npwd_sellout_listings (
...
 AVG_ROW_LENGTH = 2048,
 CHARACTER SET utf8mb4,
 COLLATE utf8mb4_general_ci;
 
+CREATE TABLE IF NOT EXISTS npwd_marketplace_reports (
```
erik-niehaus, line 87: This isn't strictly necessary but I found that when apps get to production having these two columns:

```
  `createdAt` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `updatedAt` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```

In every table add a ton of debugging/troubleshooting/features/general helpfulness. Basically when you create a new row both columns are set to the current time. If you ever perform updates to a row `updatedAt` is set to the new time.

It requires no changes to SQL queries and sort of just maintains itself.

resources/server/discord.ts
```diff
@@ -37,3 +38,28 @@ export async function reportTweetToDiscord(tweet: Tweet, reportingProfile: Profi
 
   return await axios.post(POST_CHANNEL_URL, { content }, { headers: DISCORD_HEADERS });
 }
+
+export async function reportListingToDiscord(
```
erik-niehaus, line 42: Just want to make sure this has been tested, I don't think I've run the twitter version for a few months. Although it should work.

resources/server/sv_sellout.ts
```diff
@@ -35,6 +38,35 @@ async function addListing(
...
+}
+
+async function reportListing(listingId: number, profile: string): Promise<void> {
+  const query = `INSERT INTO npwd_marketplace_reports (listing_id, profile) VALUES (?, ?)`;
+
+  await pool.query(query, [listingId, profile]);
+}
+
+async function doesReportExist(listingId: number, profile: string): Promise<boolean> {
+  const query = `SELECT * FROM npwd_marketplace_reports WHERE listing_id = ? AND profile = ?`;
```
erik-niehaus, line 63: You can `SELECT COUNT(*)` and just parse that result, it may be more performant. I'm not sure why I didn't do that on the twitter one.

Probably not super necessary since we're already filtering by id.

erik-niehaus (commented):
..

resources/import.sql
erik-sn, line 87: This isn't strictly necessary but I found that when apps get to production having these two columns:

```
  `createdAt` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `updatedAt` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```

In every table add a ton of debugging/troubleshooting/features/general helpfulness. Basically when you create a new row both columns are set to the current time. If you ever perform updates to a row updatedAt is set to the new time.

It requires no changes to SQL queries and sort of just maintains itself.

resources/server/discord.ts
erik-sn, line 42: Just want to make sure this has been tested, I don't think I've run the twitter version for a few months. Although it should work.

resources/server/sv_sellout.ts
erik-sn, line 63: You can `SELECT COUNT(*)` and just parse that result, it may be more performant. I'm not sure why I didn't do that on the twitter one.

Probably not super necessary since we're already filtering by id.

erik-sn (commented):
Just some suggestions

resources/import.sql
itschip: Just automated when I copied the whole thing kek. We should change that whole file to start on 0, right?

phone/src/apps/sellout/hooks/useListing.ts
itschip: Agree

resources/server/discord.ts
itschip: Actaully no idea 😂 I'll make sure to test it.

RockySouthpaw: Will conduct testing shortly.

RockySouthpaw: @erik-sn Issues arose with testing the twitter reporting. I DM'ed you on discord regarding it.

I think it might be best to use a webhook url instead of bot integration as webhooks already use a specified channel and are generally more secure as they don't put the entire server at risk should it get leaked. Would be less convars as well and introduce less failure points when it comes to bug testing/installation instructions. It would still need to be set in the server.cfg the same way to avoid building it though.

I can send you some resources that use it effectively as well if you need reference.

I have no issue writing good documentation for discord bot creation should webhooks not be possible.

resources/import.sql
jfrader: Nice one!

PR #143, opened 2021-03-05
[Feature] Twitter retweets

erik-sn:
This adds retweets to the twitter app:

link

Things it does:
- adds a new column in the tweets table, `retweet` that indicates if a tweet is itself a retweet of another and if so the ID of the referenced tweet
- Modifies SQL to join off of this relationship and retrieve various fields based on it
- Some logic in the UI based on whether or not we are a retweet on whether or not to show buttons, which avatar/content to show, etc.
- modifies the `Tweet` typing to handle all this

Caveats:
- I'm either too bad at SQL to get this JOIN working or this requires a second table to hold the relationship: Currently for every tweet it does not show the player if they have already retweeted something. Instead we just enforce this on the server side so people can't retweet the same tweet multiple times, and you can't retweet your own tweet.

Someone should definitely pull & test this. Appreciate any feedback.
[[Feature] Twitter retweets](https://app.gitkraken.com/glo/card/a525e8d78691420998d0fb51cea026db)


RockySouthpaw:
Works well in game. Only issue is as you said, there is no indication if someone already retweeted something. 

Currently it functions as followed:
- User will retweet. (You can't do your own.)
- User will get a notification that x person retweeted. 
- Tweet is retweeted and the icon is blue.

Once they refresh the page, the icon returns to being grey. If they try to retweet the same one, it will just turn it blue and not retweet.

I also tested the functionality of deleting a retweet and it works too.

So the only issues with this PR is the following:
- State of being retweeted isn't applied. You can probably just not store the state and get off with an error notification for if its already retweeted and prevent the icon turning blue. 🤷🏻 
- There's some prints that occur when twitter is loaded: https://i.imgur.com/p7cyrdP.png (Might just be the way we debug but i'm mentioning it anyways.

erik-sn:
Ok @RockySouthpaw I:

- added a snackbar alert when you retweet something you already have. Maybe if this ends up not being enough we can add another table to save the relationship
- removed the debugging/logging statements you noticed
- resolved merge conflicts

RockySouthpaw:
Tested the new error and it works well 👍🏻 . The icon still changes to blue as if you did retweet it though. Only issue I see with this currently.

(This PR also helped narrow down the fetching issue. Seems it takes 15 seconds after you tweet/retweet to fetch tweets again. Any tweets/retweets sent **after** that refresh are not fetched until you refresh the page. This issue has been defined better on the GitKraken.)

erik-sn:
> 
> 
> Tested the new error and it works well 👍🏻 . The icon still changes to blue as if you did retweet it though. Only issue I see with this currently.

Ok I'll add one more commit to resolve this. 

> (This PR also helped narrow down the fetching issue. Seems it takes 15 seconds after you tweet/retweet to fetch tweets again. Any tweets/retweets sent **after** that refresh are not fetched until you refresh the page. 

I have a follow up PR locally that resolves this and keeps all active players connected to the server updated to tweets.

RockySouthpaw:
If its intended behavior then the PR is tested fully and in working order. :) Just needs a code review from the others.

Great work 👍🏻

phone/src/apps/twitter/components/tweet/Tweet.tsx
```diff
@@ -86,41 +99,52 @@ export const Tweet = (tweet) => {
...
+        <div className={classes.content}>
+          <div className={classes.primary}>
+            <div className={classes.profile}>{formattedProfileName}</div>
+            <Typography
+              className={classes.date}
+              component="div"
+              variant="body2"
+              color="textSecondary"
+            >
+              {secondsToHumanReadable(t, seconds_since_tweet)}
```
TasoOneAsia, line 127: Although this wasn't altered in this PR, we should note something for the future. We do have `dayjs` installed allowing for relative time conversion from any date format. It might be a good idea to just store data with its relevant unix time rather than `seconds_since_tweet` and let dayjs do the relative conversion.

[Here](https://day.js.org/docs/en/display/from-now) is the link for that API.

resources/server/sv_twitter.ts
```diff
@@ -269,6 +287,23 @@ async function doesReportExist(tweetId: number, profileId: number): Promise<bool
...
+/**
+ * Check to see if this player has already retweeted this tweet or
+ * is the original poster of a retweeted tweet
+ * @param tweetId - ID of the tweet in question
+ * @param identifier - player identifier of the person performing the retweet
+ */
+async function doesRetweetExist(tweetId: number, identifier: string): Promise<boolean> {
+  const query = `
+  SELECT COUNT(id) as count
+  FROM es_extended.npwd_twitter_tweets
```
TasoOneAsia, line 299: We probably shouldn't have an explicit DB name here.

phone/src/apps/twitter/components/buttons/RetweetButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Nui from '../../../../os/nui-events/utils/Nui';
+
+interface IProps {
+  tweetId: number;
+  retweetId: string;
+  isRetweet: boolean | number;
+}
+
+const useStyles = makeStyles({
+  icon: {},
```
TasoOneAsia, line 15:
```suggestion
```

TasoOneAsia (changes requested):
Just a couple small things and a topic we should discuss for the future. But rest looks good 👍

phone/src/apps/twitter/components/buttons/RetweetButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+  iconFilled: {
+    color: "#00acee",
```
jfrader, line 17: There's a way implemented to use a Theme for each app.
You can see the contacts app for example which has ContactsThemeProvider and contacts.theme.ts

This way you don't need to hardcode colors and u can get them from theme.

This is really handy for example to style primary secondary buttons and textfields.

phone/src/apps/twitter/components/tweet/Retweet.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faRetweet } from '@fortawesome/free-solid-svg-icons';
+
+interface IProps {
+    profileName: string;
+}
+
+const useStyles = makeStyles({
+    root: {
+      color: '#999',
```
jfrader, line 14: Same as comment above

jfrader (approved):
Looks good, i would just do the Twitter theme provider that's missing from this app, but that can be made in another iteration and just overall remove hardcoded colors and overrides.

Nice job

phone/src/apps/twitter/components/buttons/RetweetButton.tsx
erik-sn: Ok that's a great example. There are too many custom colors/css in twitter right now to do it in this PR but I added it for the two places you have commented on. I will do a follow up PR to refactor all the custom overrides into the theme.

phone/src/apps/twitter/components/tweet/Tweet.tsx
erik-sn: Ok I can try to tackle this in a separate PR. Right now we just save the the timestamp in the database, `seconds_since_tweet` is just selected in the SQL with SQL functions.

resources/server/sv_twitter.ts
erik-sn: updated!

phone/src/apps/twitter/components/buttons/RetweetButton.tsx
erik-sn: updated!

phone/src/apps/twitter/components/buttons/RetweetButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  if (loading) {
+    return (
+      <Button disabled>
+        <CircularProgress size={22} />
+      </Button>
+    );
+  }
+
+  const className = retweeted ? classes.iconFilled : classes.icon;
```
itschip, line 55: Seems like you're missing the `icon` class.

erik-sn:
@itschip Updated. I also found out that the `yarn watch` command is not failing on TS compilation which I think I saw some discussion about in discord on the other PR. This used to work so I have to go look and see what's up. From now on I'll make sure to `yarn build` and test.

phone/src/apps/twitter/components/buttons/RetweetButton.tsx
erik-sn: updated

PR #144, opened 2021-03-05
Bugfix / Fix message bubble, notifications slideup phone and overflow and conversation scrolling

jfrader:


RockySouthpaw:
Overflow is fixed, scrolling is fixed, and notification slide up works correctly **outside** message groups.

New phone transition is great. 

Message groups also causes the issue when you open the phone for the phone to transition weirdly.

PR needs code review still.

TasoOneAsia (approved):

PR #145, opened 2021-03-05
feat(client/debug): Phone restart command

TasoOneAsia:
Can be activated by using `phone:restart` in the F8 console. This will reload the entire frame.
[feat(client/debug): Phone restart command](https://app.gitkraken.com/glo/card/4106204ed1ca494facc34e510a49cb66)

### DIFF
--- a/phone/src/Phone.tsx
+++ b/phone/src/Phone.tsx
@@ -28,6 +28,7 @@ import { usePhoneVisibility } from './os/phone/hooks/usePhoneVisibility';
 import { Snackbar } from './ui/components/Snackbar';
 import { NotificationAlert } from './os/notifications/components/NotificationAlert';
 import { useCallModal } from './os/call/hooks/useCallModal';
+import WindowSnackbar from './ui/components/WindowSnackbar';
 
 function Phone() {
   const { apps } = useApps();
@@ -53,6 +54,7 @@ function Phone() {
 
   return (
     <div>
+      <WindowSnackbar />
       <Slide direction="up" timeout={{ enter: 500, exit: 500 }} in={visibility}>
         <div className="PhoneWrapper">
           <div
--- a/phone/src/ui/components/WindowSnackbar.tsx
+++ b/phone/src/ui/components/WindowSnackbar.tsx
@@ -1,0 +1,42 @@
+import React, { useState } from 'react';
+import { Snackbar, Typography } from '@material-ui/core';
+import { useNuiEvent } from '../../os/nui-events/hooks/useNuiEvent';
+import { Alert } from './Alert';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles({
+  root: {
+    position: 'absolute',
+    width: 'auto',
+    bottom: 20,
+  },
+});
+// NOTE: Will make this more generic at some point for error handling as well
+const WindowSnackbar = () => {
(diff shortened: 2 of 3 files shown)


RockySouthpaw:
Few things:

- The current notification doesn't read well as it currently reads "Starting UI restart". While i understand we are starting the restart command, it just doesn't have good readability as it's saying We're starting the UI and then restarting it.

- It should be` "Restarting UI..."` to let the user know the UI is being restarted, and send a success notification, `"UI has been restarted."`, when the UI has been restarted as there is no indicator stating it has concluded. We are currently assuming the user is aware that once the notification goes away it's safe to open. We should be as clear and concise as possible especially since if the user open their phone while the starting notification is going off, they will need to run the command again. 

- An additional check to see if a restart is currently in progress should also be added to prevent any issues. I had an occurrence where I ran the command twice (possible), it broke the phone. Running it a 3rd time didn't solve the issue. I couldn't reproduce this occurrence though.

jfrader (approved):

TasoOneAsia:
> when the UI has been restarted as there is no indicator stating it has concluded. We are currently assuming the user is aware that once the notification goes away it's safe to open. We should be as clear and concise as possible especially since if the user opens their phone while the starting notification is going off, they will need to run the command again. 

No way to do this, when the UI restart it means it fully restarts, state is reset and the scripts are run as if they are essentially starting from scratch

> An additional check to see if a restart is currently in progress should also be added to prevent any issues. I had an occurrence where I ran the command twice (possible), it broke the phone. Running it a 3rd time didn't solve the issue and I had to restart the resource. I couldn't reproduce this occurrence though.

This would also add an additional state, that is most likely unneeded. I don't know how running it in a row could break it but ill double check

RockySouthpaw:
> > when the UI has been restarted as there is no indicator stating it has concluded. We are currently assuming the user is aware that once the notification goes away it's safe to open. We should be as clear and concise as possible especially since if the user opens their phone while the starting notification is going off, they will need to run the command again.
> 
> No way to do this, when the UI restart it means it fully restarts, state is reset and the scripts are run as if they are essentially starting from scratch
> 
Understood.
> > An additional check to see if a restart is currently in progress should also be added to prevent any issues. I had an occurrence where I ran the command twice (possible), it broke the phone. Running it a 3rd time didn't solve the issue and I had to restart the resource. I couldn't reproduce this occurrence though.
> 
> This would also add an additional state, that is most likely unneeded. I don't know how running it in a row could break it but ill double check

I couldn't reproduce the issue as I said, but you could most likely just set a variable inside the command that becomes true and wrapping the command within an if statement that only runs if said variable is false. I assume the restart will revert this back to false anyways once it restarts assuming my understanding is correct.

AvarianKnight:
> > when the UI has been restarted as there is no indicator stating it has concluded. We are currently assuming the user is aware that once the notification goes away it's safe to open. We should be as clear and concise as possible especially since if the user opens their phone while the starting notification is going off, they will need to run the command again.
> 
> No way to do this, when the UI restart it means it fully restarts, state is reset and the scripts are run as if they are essentially starting from scratch

Not completely true, you can set that the resource is restarting somewhere in the client and once the UI has finished reloading, send a NUI callback to the client and if the ui-reset var is true then show a notification saying that the phone has finished restarting.

itschip (approved):
Works and LGTM 👍

PR #146, opened 2021-03-06
fix(phone/snackbar): Fixed animations & simplified logic

TasoOneAsia:
This fixes the animations issues when transitioning in and out from an Alert. This also simplifies the useEffect logic and there isn't much of a need for a timeout + useEffect
[fix(phone/snackbar): Fixed animations & simplified logic](https://app.gitkraken.com/glo/card/9ae022fca0634990ab6e1d946146d81d)


RockySouthpaw:
LG2M 👍🏻

jfrader (approved):

PR #147, opened 2021-03-06
Feature/Twitter Notification Filters

erik-sn:
This adds notification filters for twitter, currently just two options "All Tweets" and "Only Mentions":

img

And adds filtering in the twitter notification filtering to:

1. Don't show notifications for the player's own tweets
2. If set to "Only Mentions" doesn't show notifications unless the user's profile name is a substring of the tweet message.

Caveats:

- Weird thing to note, the  `InjectDebugData` in the `TwitterApp.tsx` overwrites the profile name and will cause this "Only Mentions" to not work.
- This adds a new setting so you may have to clear local storage if running this for the first time (not sure)

[Feature/Twitter Notification Filters](https://app.gitkraken.com/glo/card/8ce1e250278d40e4b5d527966ba37d51)


phone/src/ui/hooks/useContextMenu.tsx
```diff
@@ -12,9 +12,9 @@ export const MapStringOptions = (current, onClick) => (string) => {
   };
 };
 
-export interface SettingOption {
+export interface SettingOption<T = void> {
```
erik-sn, line 15: This generic allows you to optionally specify valid values for the SettingOption

TasoOneAsia:
> Weird thing to note, the InjectDebugData in the TwitterApp.tsx overwrites the profile name and will cause this "Only Mentions" to not work.

This should never happen when the app is running with process.env.NODE_ENV === 'production'. It will do this in development though.

RockySouthpaw:
Will test in game sometime in the morning if I don’t get to it tonight. :)

jfrader (approved):
Lgtm

phone/src/Phone.tsx
```diff
@@ -48,6 +49,12 @@ function Phone() {
   useCallService();
   useDialService();
 
+  useEffect(() => {
```
jfrader, line 52: This should be in useTwitterService()

jfrader (changes requested):

RockySouthpaw:
PR isn't in a working state. For whatever reason, it's not getting the identifier when creating a profile. This doesn't exist on master or the other twitter PR.
cmd_BXu7pWF2cL

I double checked my sql and re-installed it and the issue persisted. Unsure what the actual cause of this is.

phone/src/Phone.tsx
erik-sn: Updated

erik-sn:
> This should never happen when the app is running with process.env.NODE_ENV === 'production'. It will do this in development though.

Sorry, yes I should have clarified this is in development - was just warning anyone who was doing testing it might have this behavior

> PR isn't in a working state. For whatever reason, it's not getting the identifier when creating a profile. This doesn't exist on master or the other twitter PR.

I updated `sv_main.ts` to create a profile on player load as you suggested - I wiped the tweets & profile table on my local server and I think this is resolved but I would appreciate it if you can test again. I agree that for some reason ESX is not loading before this fires, I'm not sure why.

RockySouthpaw:
I’ll test it out when I’m free, yea the whole dependency of esx player loading is a bit annoying since v1 takes a considerably longer amount of time to load in comparatively to the other ESX versions. It’s annoyingly inconvenient to code around a framework that doesn’t load in a reasonable amount of time due to it ditching core FiveM resources.

jfrader (approved):

RockySouthpaw:
image
New error.

RockySouthpaw:
Aside from the issue we disucssed, all other functionality with notifications works. Once the issue with creating profiles is fixed then all will be well.

RockySouthpaw (approved):
PR is fully functional as of the latest ESX fix.

jfrader (changes requested):

jfrader (approved):

phone/src/apps/twitter/hooks/useTwitterService.ts
```diff
@@ -64,4 +66,10 @@ export const useTwitterService = () => {
   useNuiEvent(APP_TWITTER, 'createTweetLoading', setCreateLoading);
   useNuiEvent(APP_TWITTER, 'createTweetResult', handleAddAlert);
   useNuiEvent(APP_TWITTER, 'createTweetBroadcast', setNotification);
+  
+  useEffect(() => {
```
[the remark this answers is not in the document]
erik-sn: updated

erik-sn:
Ok updated this - I think it is ready @RockySouthpaw

itschip:
I am not sure if this is just me, but there seems to be an error here. If this something that 'only works ingame', then just remove the whole emulation of that notification.

image

RockySouthpaw:
Heres a better screenshot without the weird tint chip applied smh 
Code_ttGNTpiIx8

itschip (approved):
LGTM 👍

PR #148, opened 2021-03-06
fix(messages): Sliding phone up and down when mesages apps is open

jfrader:
… fixed now


RockySouthpaw:
Was tested and is 1 line so was merged. 👍🏻

PR #149, opened 2021-03-06
updated bs.json

Hamy-os:


jfrader (approved):
Good work!

TasoOneAsia:
Needs changes to `phone/src/i18n.ts` as well for the rename and two-letter code change.

TasoOneAsia (changes requested):
Needs changes to `phone/src/i18n.ts` as well for the rename and two-letter code change.

Hamy-os:
ok done

itschip:
@Hamy-os I see there are more keys that are still in English. Would you mind changing these?

phone/src/i18n.ts
```diff
@@ -3,7 +3,7 @@ import { initReactI18next } from 'react-i18next';
 import en from './locale/en.json';
 import es from './locale/es.json';
 import fr from './locale/fr.json';
-import bh from './locale/bh.json';
+import bh from './locale/bs.json';
```
TasoOneAsia, line 6: Still incorrect, please change the name of default import and the subsequent reference as well

TasoOneAsia (changes requested):

phone/src/i18n.ts
jfrader: Omg how did I miss it lol

Also needs change in i18n.missingKeys script

Hamy-os:
be back soon

LedAndris:
smh hamy

PR #150, opened 2021-03-06
updated bh.json to be the correct lang code and translation

Hamy-os:
hello this is totally my first pr to this project


jfrader (approved):
Nice!

PR #151, opened 2021-03-06
Feat / Reusable SearchField + Fixed Messages align bottom flex with scroll (first message starts at bottom)

jfrader:
Fixed conversation bubbles alignment bottom making conflict with scrolling.

Also made a reusable ui comp. SearchField with the animation reused in messages and contacts apps


itschip (approved):
Works and looks good 👍

PR #153, opened 2021-03-06
Fix the fucking esx errors :)

jfrader:
Show initializing phone message if not ready yet
Set phone ready before sending any init events

### DIFF
--- a/phone/src/Phone.tsx
+++ b/phone/src/Phone.tsx
@@ -4,7 +4,7 @@ import './i18n';
 import { Route } from 'react-router-dom';
 import { CallModal } from './os/call/components/CallModal';
 import { HomeApp } from './apps/home/components/Home';
-import { Slide } from '@material-ui/core';
+import { Box, Paper, Slide, Typography } from '@material-ui/core';
 import { useKeyboardService } from './os/keyboard/hooks/useKeyboardService';
 import { NotificationBar } from './os/notifications/components/NotificationBar';
 import { Navigation } from './os/navigation-bar/components/Navigation';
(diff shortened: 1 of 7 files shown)


erik-sn (approved):
This is great. I think this should solve the problem well.

Do we understand why ESX takes a long time to load a player on some servers/environments? My only concern is that if the variation between users is high there could be some "bugs" where people don't get to use the phone. But that seems like an edge case.

itschip (approved):
Looks good!

RockySouthpaw:
> This is great. I think this should solve the problem well.
> 
> Do we understand why ESX takes a long time to load a player on some servers/environments? My only concern is that if the variation between users is high there could be some "bugs" where people don't get to use the phone. But that seems like an edge case.

No we dont. If the variation exist to a point of causing breaking issues between two servers with less than 30 resources then the phone will not be stable with ESX being required for functionality.

Esx V1 already has non reproducible issues with spawning players turning into micheal and is almost a year beyond its final point. We shouldn’t even be dealing with ESX issues this late in the phone quite frankly as it should of been abstracted early on.

Regardless if this PR “solves” this issue (we haven’t tested it), we would be naive to rely on some work around to fix the underlying issues that reside within ESX.

jfrader:
> > This is great. I think this should solve the problem well.
> > 
> > Do we understand why ESX takes a long time to load a player on some servers/environments? My only concern is that if the variation between users is high there could be some "bugs" where people don't get to use the phone. But that seems like an edge case.
> 
> No we dont. If the variation exist to a point of causing breaking issues between two servers with less than 30 resources then the phone will not be stable with ESX being required for functionality.
> 
> Esx V1 already has non reproducible issues with spawning players turning into micheal and is almost a year beyond its final point. We shouldn’t even be dealing with ESX issues this late in the phone quite frankly as it should of been abstracted early on.
> 
> Regardless if this PR “solves” this issue (we haven’t tested it), we would be naive to rely on some work around to fix the underlying issues that reside within ESX. 

Even tho what you say is true, these changes are needed now to continue the development knowing that we have the data we need now.

Also, these changes should be useful for when ESX is complete removed from this or at least abstracted our own data to work with, because we will still need to wait for the identifier and source to be available.

jfrader:
> This is great. I think this should solve the problem well.
> 
> Do we understand why ESX takes a long time to load a player on some servers/environments? My only concern is that if the variation between users is high there could be some "bugs" where people don't get to use the phone. But that seems like an edge case.

I don't think difference between players in loading time should matter at all.

RockySouthpaw (approved):
Tested briefly and this pr removed the error with messages pertaining to loading. Should be merged to master and then test should be conducted on https://github.com/project-error/new-phone-who-dis/pull/147 as this was another instance of the issue.

PR #155, opened 2021-03-07
feat(resources): remove all esx code, almost

itschip:
Removes almost all ESX code,.
[Feat: Remove ESX Code](https://app.gitkraken.com/glo/card/dae2624dcdcc45d9ab0077ad5fb58425)

### DIFF
--- a/resources/client/cl_main.ts
+++ b/resources/client/cl_main.ts
@@ -111,7 +111,7 @@ async function Phone(): Promise<void> {
 }
 
 // triggerd when the player is ready
-onNet(events.PHONE_IS_READY, fetchOnInitialize);
+onNet(events.PLAYER_IS_READY, fetchOnInitialize);
 
 AddEventHandler('onResourceStop', function (resource: string) {
   if (resource === GetCurrentResourceName()) {
--- a/resources/client/cl_main.ts
+++ b/resources/client/cl_main.ts
@@ -5,13 +5,6 @@ import { phoneCloseAnim, phoneOpenAnim, removePhoneProp } from './functions';
 let isPhoneOpen = false;
 let isPhoneReady = false;
 
-while (true) {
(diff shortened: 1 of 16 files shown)


TasoOneAsia (changes requested):

TasoOneAsia:
Needs to be reviewed by others

RockySouthpaw:
So i tested this on my V1 Final server and here is the summary: 

- Number generating **only** works if they have a name already. 
     - On user first join they don't have a name which we try to fetch. Solution is to just remove this and make users 
- Theres some issues getting identifier when you open the phone. 
     -  https://cdn.discordapp.com/attachments/791854455468458042/818720637576282112/unknown.png
- Calling doesn't work currently at all as a result of the above issue.
- Theres an error as well when quiting while the server is up.

I made taso aware of the issues so its all in discord: Can follow it from here: https://discord.com/channels/791854454760013827/791854455468458042/818720640282394634

We will also need to include in some place that they need to be on at least `Artifact Version 3598`. I'll add to the Docs at some point.

TasoOneAsia:
Fixed @RockySouthpaw

RockySouthpaw:
@TasoOneAsia Will retest now.

RockySouthpaw:
> * Number generating **only** works if they have a name already.
**Fixed** 
>   * On user first join they don't have a name which we try to fetch. Solution is to just remove this and make users
**Fixed** 
> * Theres some issues getting identifier when you open the phone.
**Fixed** 
> * Calling doesn't work currently at all as a result of the above issue.
**Fixed** 
> * Theres an error as well when quiting while the server is up.
**Fixed** 

PR is ready to be merged. 👍🏻

RockySouthpaw (approved):

TasoOneAsia:
Will not be squashed/rebased as some commits have important comments

PR #161, opened 2021-03-10
Fix remaining issues for messages

itschip:
Some stuff still missing.
[Fix remaining issues for messages](https://app.gitkraken.com/glo/card/ce6e8fbc44b84b87a2dcb2b39e5d14c5)

### DIFF
--- a/phone/src/apps/messages/components/modal/MessageModal.tsx
+++ b/phone/src/apps/messages/components/modal/MessageModal.tsx
@@ -117,6 +117,9 @@ export const MessageModal = () => {
 
   const phoneNumbers = activeMessageGroup?.phoneNumbers || [];
 
+  const targetNumber = phoneNumbers.find((n) => myNumber !== n);
+  console.log('TARGET NUMBER: ', targetNumber);
+
   return (
     <div>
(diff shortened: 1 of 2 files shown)


phone/src/apps/messages/components/modal/MessageModal.tsx
```diff
@@ -117,6 +117,9 @@ export const MessageModal = () => {
 
   const phoneNumbers = activeMessageGroup?.phoneNumbers || [];
 
+  const targetNumber = phoneNumbers.find((n) => myNumber !== n);
+  console.log('TARGET NUMBER: ', targetNumber);
```
jfrader, line 121: Remove console log

jfrader (changes requested):

phone/src/apps/messages/components/modal/MessageModal.tsx
itschip: Yes

RockySouthpaw:
I suggest merging this to master as a fix for the issue around checking if contact exist as this fixes it but there are 11 other issues in total and would be deathremential to tackle all in 1 PR given the changes to  master that occur so frequently.

PR #163, opened 2021-03-10
All new objects translated

synterrr:


TasoOneAsia (commented):
LGTM fixes missing objects

PR #167, opened 2021-03-14
Refactor/events enums

TasoOneAsia:
Switches net events in cl/sv to using enums rather than the events object.


jfrader (commented):
Could you replace Nui.send in frontend as well?

erik-sn (approved):
This is great, I had something like this done for Twitter individually but doing them all in one go is best.

Have you thought about prepending every event with `npwd:` (like `npwd:myCoolEvent`) just so it is namespaced and there will never be a possibility of collision with another package?

jfrader (approved):

TasoOneAsia:
> Have you thought about prepending every event with `npwd:` (like `npwd:myCoolEvent`) just so it is namespaced and there will never be a possibility of collision with another package?

Yea we can do that, should be pretty easy now with the enums.

Also @erik-sn & @jfrader just pushed a commit with event enums attached to their respective Nui.send counterparts

TasoOneAsia:
> Have you thought about prepending every event with `npwd:` (like `npwd:myCoolEvent`) just so it is namespaced and there will never be a possibility of collision with another package?

Just prefixed all events with the last commit

TasoOneAsia:
Note this does break with `yarn start` as react-scripts is expecting all the modules to be within `./src` folder. This can be fixed with a couple of changes.

itschip (approved):
lgtm

TasoOneAsia:
@jfrader @erik-sn  @itschip  If we are all good, I want to land this PR on master. So if you have any final comments, let me know

PR #168, opened 2021-03-17
feat(locale): Added a locale setting in Settings app

TasoOneAsia:
This removes the need to rebuild the phone in order to change language

PR #169, opened 2021-03-17
tweak(dialer): Several fixes and improvements

TasoOneAsia:
Time has been changed to a relative format. Locale specifics are pending
merge of #168.

Several typos have been fixed where "DAILER" was referenced instead of
"DIALER"

A translation key was added for when a user has no call history.

The input box has been unlocked so users can use their keyboard directly
rather than using the dial pad.

### DIFF
--- a/phone/src/apps/dialer/components/DialerApp.tsx
+++ b/phone/src/apps/dialer/components/DialerApp.tsx
@@ -28,7 +28,7 @@ export const DialerApp = () => {
             <Route exact path="/phone">
               <DialerHistory calls={history} />
             </Route>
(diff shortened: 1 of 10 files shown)


RockySouthpaw:
Can you add to this PR to auto populate the dash as well? Depending on the config number.

TasoOneAsia:
> Can you add to this PR to auto populate the dash as well? Depending on the config number.

This is dependent on how we decide how to treat the phone number from server -> NUI. So probably best to hold off with that for the moment.

itschip (approved):

PR #170, opened 2021-03-17
Feature/default profile names

erik-sn:
This more elegantly handles the case of generating twitter profile names, with any eye towards using it for tinder profile names too.

There are two primary configuration options we're concerned about:

1. `twitter.generateProfileNameFromUsers` - whether or not we automatically generate profile names
2. `twitter.allowEdtiableProfileName` - whether or not we allow players to edit their own profile names

**Various cases:**

- When 1 is true we attempt to generate profile names based on:

1. `firstname_lastname` if both exist
2. only `firstname` if it exists
3. only `lastname` if it exists
4. `phone_number` otherwise

- When 1 is false and 2 is true we show an input field and allow the user to type in a profile name of their choosing
- When 1 is false AND 2 is false we show a select field populated by name (same logic as above) AND phone number:

img

In the future specifying which table/columns to read these values from will make this much more flexible.


phone/src/apps/messages/components/modal/MessageModal.tsx
```diff
@@ -157,7 +157,7 @@ export const MessageModal = () => {
                 </Button>
               ) : null}
               {!activeMessageGroup?.isGroupChat &&
-              getDisplayByNumber(targetNumber) == targetNumber ? (
+              getDisplayByNumber(targetNumber) === targetNumber ? (
```
erik-sn, line 160: random thing my linter highlighted

resources/server/players/sv_players.ts
```diff
@@ -89,6 +89,54 @@ export async function getPlayerInfo(identifier: string): Promise<PlayerQueryResp
   return { firstname, lastname, phone_number };
 }
 
+const clean = (input: string) => (input ? input.replace(/[^0-9a-z]/gi, '') : input);
```
erik-sn, line 92: If anyone has some elite JS to simplify/consolidate/condense all these functions let me know!

resources/server/sv_twitter.ts
```diff
@@ -316,12 +316,41 @@ onNet(events.TWITTER_GET_OR_CREATE_PROFILE, async () => {
...
+      emitNet(events.TWITTER_GET_OR_CREATE_PROFILE_NULL, _source, defaultProfileNames);
+    } else {
+      emitNet(events.TWITTER_GET_OR_CREATE_PROFILE_SUCCESS, _source, profile);
+    }
   } catch (e) {
     twitterLogger.error(`Failed to get or create profile, ${e.message}`, {
       source: _source,
     });
-    emitNet(events.TWITTER_GET_OR_CREATE_PROFILE_FAILURE, _source);
+    emitNet(events.TWITTER_GET_OR_CREATE_PROFILE_FAILURE, _source, null);
```
TasoOneAsia, line 333: If there is a failure is there any reason we have to emit null? Can't it be assumed from a failure event?

resources/server/players/sv_players.ts
```diff
@@ -89,6 +89,54 @@ export async function getPlayerInfo(identifier: string): Promise<PlayerQueryResp
...
+
+  if (firstname && lastname) {
+    return `${firstname}_${lastname}`;
+  } else if (firstname) {
+    return firstname;
+  } else if (lastname) {
+    return lastname;
+  } else if (phone_number) {
+    return phone_number;
+  }
```
TasoOneAsia, line 118: Would be a tiny bit more efficient as a switch but its inconsequential in this case.

TasoOneAsia (approved):
screenshot

Just some general linter warnings as well but lgtm 👍

erik-sn:
@TasoOneAsia good points, updated with your feedback

itschip (approved):

PR #172, opened 2021-03-17
feat(phone/settings): Schema validation & Reset Settings Action

TasoOneAsia:
You can now reset your settings to default using the action in the Settings App

Setting schema validation has been added that will temporarily reset to default settings if
the JSON string contains unexpected values. A alert will pop up informing the user
that they should reset to default settings using the action in the settings app,
as a permanent solution.


jfrader (approved):

itschip (approved):

itschip:
Fix conflicts

PR #175, opened 2021-03-22
Change license to GPL

jfrader:
The main point of this license change is that we have a wide community of developers to backup how a particular license would work, instead of using a license that is not used and doesnt have warrantee clauses for the developers, while ensuring that any modification made to this project is subject to keep the same open source and freedom ideals.

Useful links:
- https://www.gnu.org/licenses/gpl-faq.html

Please give your opinion/feedback/comments in this PR

EDIT: As discussed a bit, a better proposal would be GPL instead of AGPL to avoid conflict of interest between phone and other resources.

Some of the points discussed offline:
- Non-commercial use is a vague term which open source community avoids to restrict, it can cause legal problems to any of the contributors as the definiton is not concrete.
- There are no licenses in the open source community that can help on this aspect of non-commercial, and the ones that exist are not recommended for software.
- AGPL restricts other resources on their license if they directly connect via a network message to the phone that's why this proposal is changed to be only GPLv3


LICENSE
```diff
@@ -1,651 +1,595 @@
...
-solutions will be better for different programs; see section 13 for the
-specific requirements.
-
-You should also get your employer (if you work as a programmer) or school,
-if any, to sign a “copyright disclaimer” for the program, if necessary.
-For more information on this, and how to apply and follow the GNU AGPL, see
+If the program does terminal interaction, make it output a short notice like this
+when it starts in an interactive mode:
+
+    <program>  Copyright (C) <year>  <name of author>
```
AvarianKnight, line 576: needs to be filled in

AvarianKnight (changes requested):

LICENSE
jfrader: Thanks!

jfrader:
> 
> 
> Wow, writing yourself first..smh :kappa:

LMAO

RockySouthpaw:
The license names should be sorted alphabetically. I also don’t understand how GitHub usernames express copyright ownership ether 🤷‍♀️ it’s probably something that should contain real names otherwise there is likely no legal standing.

jfrader:
@RockySouthpaw @itschip 
I left the order that was before and infered that it was in order of commitment. I don't care about it, so go and put the order that you prefer.

jfrader:
About the other subject @RockySouthpaw I don't think names would make much difference because I think the only thing that matters is the verification in github or your email in the commits but I'm not sure. Maybe it's best to add the real names.

All of these changes can be made quickly but I don't want to move the focus of the conversation here, although I still think your points are important too.

PR #176, opened 2021-03-22
feat(phone): Added a game time indicator to top bar

TasoOneAsia:
image

Adds the current game time to the top bar, the game time is sent every time the phone is opened. This can also be done in a set interval, but I have commented that for now.

### DIFF
--- a/phone/src/os/notifications/components/NotificationBar.tsx
+++ b/phone/src/os/notifications/components/NotificationBar.tsx
@@ -16,6 +16,7 @@ import ArrowDropUpIcon from '@material-ui/icons/ArrowDropUp';
 import Default from '../../../config/default.json';
 import { useNotifications } from '../hooks/useNotifications';
 import { NotificationItem } from './NotificationItem';
+import usePhoneTime from '../../phone/hooks/usePhoneTime';
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -74,6 +75,8 @@ export const NotificationBar = () => {
     setBarUncollapsed,
   } = useNotifications();
 
+  const time = usePhoneTime();
+
   return (
     <>
       <Grid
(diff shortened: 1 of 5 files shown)


RockySouthpaw:
LG2M. Maybe a time format option instead of forcing the use of 24 hour time. 🤷🏻 only thoughts really.

AvarianKnight:
Does this take into account if people decide to change the default game time MS? You can use [GetMillisecondsPerGameMinute](https://docs.fivem.net/natives/?_0x2F8B4D1C595B11DB) to check if they modified.

TasoOneAsia:
> Does this take into account if people decide to change the default game time MS? You can use [GetMillisecondsPerGameMinute](https://docs.fivem.net/natives/?_0x2F8B4D1C595B11DB) to check if they modified.

There isn't live update intervals, I commented that part out for this exact reason. For the moment the game time is only sent when the phone is opened.

TasoOneAsia:
> LG2M. Maybe a time format option instead of forcing the use of 24 hour time. 🤷🏻 only thoughts really.

I would need to add a whole bunch more time conversions functions for this 😢

AvarianKnight:
> > Does this take into account if people decide to change the default game time MS? You can use [GetMillisecondsPerGameMinute](https://docs.fivem.net/natives/?_0x2F8B4D1C595B11DB) to check if they modified.
> 
> There isn't live update intervals, I commented that part out for this exact reason. For the moment the game time is only sent when the phone is opened.

Can you not just send the update interval every time the phone opens?

TasoOneAsia:
I can, its just the interval spams our debug logs hard, so I don't wanna implement it just yet.

RockySouthpaw (approved):
Will need to make it up date in real time later on but tested and works fine.

PR #177, opened 2021-03-22
Feat(phone): Update to use all enums. 

itschip:


phone/src/apps/bank/hooks/useBankService.ts
```diff
@@ -21,8 +22,8 @@ export const useBankService = () => {
     });
   };
 
-  useNuiEvent('BANK', 'setTransaction', setTransaction);
-  useNuiEvent('BANK', 'setCredentials', setCredentials);
+  useNuiEvent('BANK', BankEvents.SEND_TRANSFERS, setTransaction);
+  useNuiEvent('BANK', BankEvents.SEND_CREDENTIALS, setCredentials);
   useNuiEvent('BANK', 'setAlert', handleAddAlert);
```
jfrader, line 27: please replace all the events

itschip: You....you just said...alright.

resources/client/cl_bank.ts
```diff
@@ -35,7 +36,7 @@ onNet(BankEvents.TRANSACTION_ALERT, (result: IAlertProps) => {
   SendNuiMessage(
     JSON.stringify({
       app: 'BANK',
-      method: 'setAlert',
+      method: PhoneEvents.SET_ALERT,
```
jfrader, line 39: Dont use a global one if we are not changing the UI too, because we will be handling the same alert once per app repeated! Use scoped for now

phone/src/apps/bank/hooks/useBankService.ts
TasoOneAsia: These methods won't work in the useNuiEvent, as the appName factory fails to parse them correctly if they have a prefix of `npwd`

jfrader: they already had a colon so it should be fine tho

jfrader: already fixed messages like this @TasoOneAsia

jfrader (approved):

PR #179, opened 2021-03-24
Feature/tinder 

erik-sn:
This is basically Tinder, called `match` in the code. Name/icon or w/e else is flexible.

[Demo of the app](https://streamable.com/43m0cs)

- Two tables (profiles & views)
- UI is pretty straightforward, 3 "views": swiping, matches and profile editor
- I had previous implemented some crazy matchmaking algorithm that I spent way too much time on but I eventually concluded I didn't know enough about it and didn't think the populations of fivem servers (16-~200?) warranted anything more complex than a naive "get all profiles" kind of deal.
- This also means there is no concept of sex/gender involved in this, you are basically presented with everyone

This should be definitely be pulled/tested - you need at least two players to do so. [I wrote up this test plan](https://docs.google.com/document/d/114M1_2TBKiZ8K7vumTWoY_MQcmTogE59cOJmowcE7MI/edit) to help me do final testing on my side.


phone/src/apps/match/hooks/useMatchService.ts
```diff
@@ -0,0 +1,86 @@
...
+    handleAddAlert(alert);
+    setErrorLoadingMatches(true);
+  };
+
+  const _handleUpdateProfile = (profile: Profile) => {
+    _setProfile(profile);
+    handleAddAlert({ message: 'APPS_MATCH_UPDATE_PROFILE_SUCCEEDED', type: 'info' });
+  };
+
+  useNuiEvent('MATCH', MatchEvents.GET_PROFILES_SUCCESS, _setProfiles);
```
erik-sn, line 73: @jfrader I had implemented all this before you mentioned the `useNuiCallback` but I want to revisit that because it seems really useful.

typings/match.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface FormattedMatch extends FormattedProfile {
+  matchedAtFormatted: string;
+}
+
+export interface Like {
+  id: number; // profile id
+  liked: boolean;
+}
+
+export enum MatchEvents {
```
erik-sn, line 40: This is great @TasoOneAsia thanks for doing that refactor

RockySouthpaw:
Thank you for providing test plan!

RockySouthpaw:
I think leaving out sex/sexual orientation is a good idea and I'm glad you did this given the sex ratio on fivem and IRL dating apps.

It may be a good idea to add a way to un-match someone but not needed right now at all.

**Small Suggestions**
- It may be a good idea to render a text stating why the profile was failed to create should you try to create. Should you try to use a profile name that already exist, you get met with an error, but its non descriptive.
- Adding profile pic from gallery would be great for this as well as other apps tbh.

Other than the suggestion, everything worked fine for me. I created multiple fake profiles and everything went well.

RockySouthpaw (approved):

RockySouthpaw:
Tested notifications, i think the way it does it now is perfectly fine as it doesnt trigger a normal notification, only when you open the app.

Good work 👍🏻

PR #184, opened 2021-03-26
Improve/marketplace

itschip:


phone/src/os/apps/config/apps.tsx
```diff
@@ -149,8 +149,8 @@ export const APPS: IAppConfig[] = [
     notificationIcon: <FontAwesomeIcon icon={faAd} fixedWidth size="xs" />,
     backgroundColor: SELLOUT_APP_PRIMARY_COLOR,
     color: SELLOUT_APP_ICON_COLOR,
-    path: '/sellout',
-    Route: () => <AppRoute id="SELLOUT" path="/sellout" component={SelloutApp} />,
+    path: '/marketplace',
+    Route: () => <AppRoute id="SELLOUT" path="/marketplace" component={MarketplaceApp} />,
```
jfrader, line 153: id here

RockySouthpaw:
Couple things:
- The description field still has scrolling side ways if the text isn't broken up. (Twitter had same issue at one point but was fixed.)
Example: 
`aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`.

- The transition to home works but the icon doesn't have its color changed. (Shows home as grey despite it being focused.)

- The notification covers the bottom buttons. May need to be moved up :
image


**Minor enchancement**
- The max character is good to add. I would suggest triggering some sort of notification or text saying the max character has been reached though to avoid people thinking its bugged.

Aside from the above, everything else works well. 👍🏻

phone/src/os/apps/config/apps.tsx
itschip: Thanks! Forgot this one

PR #185, opened 2021-03-30
tweak(phone/settings): simplify custom background

itschip:


phone/src/Phone.tsx
```diff
@@ -104,9 +104,10 @@ function Phone() {
               id="phone"
               className="PhoneScreen"
               style={{
-                backgroundImage: settings.customWallpaper
-                  ? `url(${settings.customWallpaper})`
-                  : `url(media/backgrounds/${settings.wallpaper.value})`,
+                backgroundImage:
+                  settings.wallpaper.label === 'Custom Wallpaper'
```
jfrader, line 108: what happens in other languages?

phone/src/apps/settings/components/SettingsApp.tsx
```diff
@@ -73,9 +72,12 @@ export const SettingsApp = () => {
   };
 
   const wallpapers = config.wallpapers.map(
-    MapSettingItem(settings.wallpaper, (val: SettingOption) =>
-      handleSettingChange('wallpaper', val),
-    ),
+    MapSettingItem(settings.wallpaper, (val: SettingOption) => {
+      if (val.label === 'Custom Wallpaper') {
```
jfrader, line 76: same, what happens in other language?

phone/src/Phone.tsx
itschip: Well...that was the issue that I was preparing you for, since the 'Custom Wallpaper' is in the default.json file, I need to figure out a way to fix it, or the users would have to change this manually. I am not quite sure how to do this, or atleast not the best way. So I want some input on this.

phone/src/apps/settings/components/SettingsApp.tsx
itschip: See comment in first commit.

phone/src/Phone.tsx
```diff
@@ -104,9 +104,10 @@ function Phone() {
               id="phone"
               className="PhoneScreen"
               style={{
-                backgroundImage: settings.customWallpaper
-                  ? `url(${settings.customWallpaper})`
-                  : `url(media/backgrounds/${settings.wallpaper.value})`,
+                backgroundImage:
+                  settings.wallpaper.label === t('APPS_SETTINGS_OPTIONS_CUSTOM_WALLPAPER')
```
TasoOneAsia, line 108: This creates an edge case where if the user changes their language after setting a wallpaper, the translation will not *match* the `settings.wallpaper.label` in localStorage. This is why I had previously suggested the utility function for determining whether a wallpaper value was a default or not.

A small hack for this would be to detect whether the `settings.wallpaper.value` is a url, as the default wallpapers will not store a URL. Not ideal but it would work if you don't want to use the utility function I suggested.

Edit: This is assuming you meant to make the label a translation key in the setter as well in `WallpaperModal.tsx`

phone/src/apps/settings/components/SettingsApp.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
```
TasoOneAsia, line 1: This isn't referenced anywhere, yeet it.

```suggestion
import React, { useState } from 'react';
```

phone/src/apps/settings/components/WallpaperModal.tsx
```diff
@@ -37,11 +37,8 @@ export default function WallpaperModal() {
   };
 
   const handleNewWallpaper = () => {
-    handleSettingChange('customWallpaper', value);
-  };
-
-  const handleRemoveWallpaper = () => {
-    handleSettingChange('customWallpaper', '');
+    handleSettingChange('wallpaper', { label: 'Custom Wallpaper', value });
```
TasoOneAsia, line 40: Here, this isn't being set as translation key, but it is being referenced as a translation key in Phone.tsx.

TasoOneAsia (changes requested):
Also a couple small suggestions for the modal itself

modal

Add a title and reduce the width of the container div. Reducing empty space will make it look nicer. Also adding a cancel button instead of the close will help populate the space with the other action, resulting in a better looking modal.

Here is an quick example I made in Figma:

dialog

phone/src/apps/settings/components/WallpaperModal.tsx
itschip: xd moment

phone/src/apps/settings/components/WallpaperModal.tsx
```diff
@@ -21,59 +21,43 @@ export default function WallpaperModal() {
...
-        >
-          {t('APPS_SETTINGS_SET_CUSTOM_WALLPAPER')}
-        </Button>
-      </Box>
-    </Modal>
+    <DialogForm
+      open={customWallpaperModal}
+      handleClose={() => setCustomWallpaperModal(false)}
+      onSubmit={handleNewWallpaper}
+      title="Custom wallpaper"
```
TasoOneAsia, line 52: Will need translation keys for all these boys

phone/src/apps/settings/components/WallpaperModal.tsx
```diff
@@ -21,59 +21,43 @@ export default function WallpaperModal() {
...
   const handleSettingChange = (key: string | number, value: unknown) => {
     setSettings({ ...settings, [key]: value });
   };
 
   const handleNewWallpaper = () => {
-    handleSettingChange('wallpaper', { label: 'Custom Wallpaper', value });
-    setCustomWallpaperModal(false);
+    if (isImageAndUrl(value)) {
+      handleSettingChange('wallpaper', {
+        label: t('APPS_SETTINGS_OPTIONS_CUSTOM_WALLPAPER'),
```
TasoOneAsia, line 38: If we set just the translation key here rather than the actual translation, we can just directly translate it on both ends rather than storing a hardcoded locale.

phone/src/locale/en.json
```diff
@@ -192,6 +192,7 @@
     "APPS_SETTINGS_OPTION_NOTIFICATIONVOL": "Notification Volume",
     "APPS_SETTING_STREAMER_MODE_DESC": "Hide images before they are shown",
     "APPS_SETTINGS_SET_CUSTOM_WALLPAPER": "Set new wallpaper",
+    "APPS_SETTINGS_INVALID_WALLPAPER": "Not a valid image or link!",
```
TasoOneAsia, line 195: If we r gonna keep with this locale style should I revert my nested key format?

phone/src/apps/settings/components/WallpaperModal.tsx
itschip: Yep, was thinking about that.

phone/src/locale/en.json
itschip: Ehh...I can rather change it. Nested looks better.

RockySouthpaw:
Will need to test this branch to see if it fixes it/doesn’t have the issue but on master, if the URL is long enough it will add a scroll bar to the settings app which looks absurd. 

I haven’t had time to test this branch so I’m unsure if this one even has that issue but I’m making a note of it here mostly as a self reminder. A good url to test would be a discord image.

RockySouthpaw:
Should add the custom wallpaper to the top as well as sort the rest of the wallpapers so they're alphabetical. 

image

phone/src/apps/settings/components/WallpaperModal.tsx
TasoOneAsia: Still needs changing

phone/src/ui/components/DialogForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+  },
+  displayNone: {
+    display: 'none',
+  },
+});
+
+interface DialogFormProps {
+  children: React.ReactNode;
+  open: boolean;
+  handleClose: any; // No idea what those types are
```
TasoOneAsia, line 32: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void

TasoOneAsia (changes requested):

PR #187, opened 2021-03-31
fix(sql): don't include extra SQL data

AvarianKnight:
Fixes error with coalition not existing


jfrader (approved):
This was needed, LGTM.

PR #190, opened 2021-04-01
RFR: Adjust identifier length + add backticks.

RockySouthpaw:
Made the default identifier length `48` since this is the maximum length if `license:` isn't removed and is something those using `esx_kashacters` will already have to do. I also adjusted the instances where it was `255` for some reason so those are `48` as well.

The addition of back ticks is just for formatting. The readability is much better inside VS and sql editors as the text is highlighted as a result. Its's also how most of the SQL was anyways.


RockySouthpaw:
Items are also removed because phone currently doesn't have that functionality and the way they were added assumes the structure of the items table.

PR #198, opened 2021-04-04
feat(resources/main): disable controls

itschip:
CPU msec goes brrrr

### DIFF
--- a/resources/client/cl_main.ts
+++ b/resources/client/cl_main.ts
@@ -11,6 +11,7 @@ import { BankEvents } from '../../typings/bank';
 import { PhotoEvents } from '../../typings/photo';
 import { CallEvents } from '../../typings/call';
 import { MatchEvents } from '../../typings/match';
+import { Delay } from '../utils/fivem';
 
 let isPhoneOpen = false;
 let isPhoneReady = false;
@@ -51,6 +52,7 @@ const getCurrentGameTime = () => {
 
 const showPhone = async (): Promise<void> => {
   isPhoneOpen = true;
+  disableControls();
   const time = getCurrentGameTime();
   await phoneOpenAnim(); // Animation starts before the phone is open
   emitNet('phone:getCredentials');
@@ -58,6 +60,7 @@ const showPhone = async (): Promise<void> => {
   sendMessage('PHONE', 'setVisibility', true);
   sendMessage('PHONE', 'setTime', time);
   SetNuiFocus(true, true);
+  SetNuiFocusKeepInput(true);
 };
 
 const hidePhone = async (): Promise<void> => {
@@ -65,6 +68,7 @@ const hidePhone = async (): Promise<void> => {
   sendMessage('PHONE', 'setVisibility', false);
   await phoneCloseAnim();
   SetNuiFocus(false, false);
+  SetNuiFocusKeepInput(false);
 };
 
 /* * * * * * * * * * * * *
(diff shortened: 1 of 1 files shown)


RockySouthpaw:
It would probably be best to disable all control actions and enable a select amount instead. Controls 32-35 (WASD) for movement, 249 (N) for push to talk, and 23 (F)for entering a vehicle. 

With the current configuration assuming I’m reading it correctly, It looks like you’re still able to open the weapon wheel, punch (would probably happen on left clicks), shoot, and other actions that probably shouldn’t be possible while holding the phone. 

You would then just need to enable all controls again once you’re done. 

I can conduct some testing later and make sure but just from looking over the code and my own testing previously, these evwnts will most likely occur.

RockySouthpaw:
Tested in game and it's as described above.

https://user-images.githubusercontent.com/55056068/113645626-332e7000-9655-11eb-93c4-53b2191764c0.mp4

Will need to go about this another way than you are currently.

RockySouthpaw:
Yea it looks like once you do `DisableAllControlActions(0);` you cant enable them from then on 🤔 So the best way would be an obnixious serious of disabling controls lol

resources/client/cl_main.ts
```diff
@@ -196,3 +200,30 @@ on(`__cfx_nui:${PhoneEvents.CLOSE_PHONE}`, async (data: any, cb: Function) => {
...
+      if (!isPhoneOpen) {
+        clearTick(loopTick);
+        return;
+      }
+
+      if (isPhoneOpen) {
+        DisableControlAction(0, 1, true);
+        DisableControlAction(0, 2, true);
+        DisableControlAction(0, 16, true);
+        DisableControlAction(0, 17, true);
```
RockySouthpaw, line 218:
```suggestion
        DisableControlAction(0, 17, true); //    Select Previous Weapon
        DisableControlAction(0, 22, true); //    Jump
        DisableControlAction(0, 24, true); //    Attack
        DisableControlAction(0, 25, true); //    Aim 37
        DisableControlAction(0, 37, true); //    Weapon Wheel
```

resources/client/cl_main.ts
```diff
@@ -196,3 +200,30 @@ on(`__cfx_nui:${PhoneEvents.CLOSE_PHONE}`, async (data: any, cb: Function) => {
...
+
+  const loopTick = setTick(
+    async (useTimeout = false): Promise<void> => {
+      if (!isPhoneOpen) {
+        clearTick(loopTick);
+        return;
+      }
+
+      if (isPhoneOpen) {
+        DisableControlAction(0, 1, true);
```
RockySouthpaw, line 215:
```suggestion
        DisableControlAction(0, 1, true);  //    Look Left Right
```

resources/client/cl_main.ts
```diff
@@ -196,3 +200,30 @@ on(`__cfx_nui:${PhoneEvents.CLOSE_PHONE}`, async (data: any, cb: Function) => {
...
+        DisableControlAction(0, 2, true);
```
RockySouthpaw, line 216:
```suggestion
        DisableControlAction(0, 2, true);  //    Look Up Down
```

resources/client/cl_main.ts
```diff
@@ -196,3 +200,30 @@ on(`__cfx_nui:${PhoneEvents.CLOSE_PHONE}`, async (data: any, cb: Function) => {
...
+        DisableControlAction(0, 16, true);
```
RockySouthpaw, line 217:
```suggestion
        DisableControlAction(0, 16, true); // 	 Select Next Weapon
```

RockySouthpaw:
@TasoOneAsia can this be closed since it’s being done differently now? I assume the current edits are essentially null and void.

TasoOneAsia:
Yes can be closed as this and any other frame ticks should probably be handled in lua to keep that precious resource time number down.

PR #202, opened 2021-04-08
Fix fetch on initialize (fixing notifications too)

jfrader:
Player Ready also needs to know if NUI is loaded seems like, as events are not being received at that point, so now the two moments race and whatever is first will fetchOnInit


jfrader:
This should fix the issue with notifications when phone wasnt open yet @RockySouthpaw

jfrader:
Also adds multiline support to messages input field @RockySouthpaw

itschip (approved):

PR #212, opened 2021-04-16
Keep input while phone is open

itschip:


futile.lua
```diff
@@ -0,0 +1,32 @@
+isPhoneOpen = false
```
RockySouthpaw, line 1: You don’t need to define the variable

futile.lua
```diff
@@ -0,0 +1,32 @@
...
+
+CreateThread(function()
+    while true do
+        Wait(5)
```
RockySouthpaw, line 5: This has to be every frame. Especially on slow PC’s.
```suggestion
        Wait(0)
```

futile.lua
```diff
@@ -0,0 +1,32 @@
...
+            DisableControlAction(0, 1, true) -- Look Left/Right
+            DisableControlAction(0, 2, true) -- Look up/Down
+            DisableControlAction(0, 16, true) -- Next Weapon
+            DisableControlAction(0, 17, true) -- Select Previous Weapon
+            DisableControlAction(0, 22, true) -- Jump
+            DisableControlAction(0, 24, true) -- Attack
+            DisableControlAction(0, 25, true) -- Aim
+            DisableControlAction(0, 37, true) -- Weapon Wheel
+        else
+            EnableControlAction(0, 1, true) -- Look Left/Right
```
RockySouthpaw, line 16: You actually shouldn’t need to enable the control. You’re disabling them in a loop so as long as that loop isn’t running, there is no need to enable them every 1000ms. This current way will likely cause issues with death scripts too.

RockySouthpaw:
Should also rename the file from `futile.lua` to `controlActions.lua` or anything that will denote what it is 🤷🏻‍♀️

fxmanifest.lua
```diff
@@ -3,8 +3,9 @@ game "gta5"
 description 'js runtime monkaW'
 authors { "itschip",  "erik-sn", "TasoOneAsia", "kidz", "RockySouthpaw"}
 
-client_script {
-    'resources/dist/client/*.client.js'
+client_scripts {
+    'resources/dist/client/*.client.js',
+    'futile.lua'
```
RockySouthpaw, line 8: You could just glob it as `*.lua` I guess and keep the meme name 🤷🏻‍♀️ but this is likely the only lua file to exist in the entire repo so.

futile.lua
RockySouthpaw: Actually you do because of `isPhoneOpen = not isPhoneOpen`

You could of just done the whole if true or false stuff but your way seems cleaner. 
I would of passed the arg to the event to denote if it’s enable or disabled for easier refactoring but to each their own.

futile.lua
RockySouthpaw: 1000ms might be too long if someone closes the phone and tries to reopen it. Would need to test it

futile.lua
```diff
@@ -0,0 +1,32 @@
...
+            EnableControlAction(0, 2, true) -- Look up/Down
+            EnableControlAction(0, 17, true) -- Select Previous Weapon
+            EnableControlAction(0, 22, true) -- Jump
+            EnableControlAction(0, 24, true) -- Attack
+            EnableControlAction(0, 25, true) -- Aim
+            EnableControlAction(0, 37, true) -- Weapon Wheel
+            EnableControlAction(0, 16, true) -- Next Weapon
+            Wait(1000)
```
TasoOneAsia, line 24: Remove all EnableControlAction

TasoOneAsia (changes requested):
Rest looks goods, with Rocky's comments in mind as well

RockySouthpaw:
Needs some refactoring as closing the phone via the home doesn't change the state so if you open the phone again, it will not disable any control actions.

Changing to draft.

RockySouthpaw:
# Known Issues
- If a server has any menu bound to any key that is typically used when writing a message (A-Z, commas, periods, ect) then that menu will pop up. This is a decent issue if you have `vMenu` or a similiar menu bound to one of these as closing it will also close the phone essentially making typing impossible. Majority of servers use M for `vMenu.`

- People will move slightly every time they type the letters `W,A,S,D` as we do not freeze them. This just looks unnatural as they would end up doing 360 degree turns every sentence while also stepping into traffic. You would essentially only be able to text if you're in a corner or in the passenger seat.

These are the only things I've come across so far. I disabled most the controls I could think of while testing.

The only solution I have would be to disable all control actions if they're clicking on a text field or `SetNuiFocusKeepInput(false)` within text boxes.

resources/client/cl_controls.lua
```diff
@@ -0,0 +1,44 @@
...
+            DisableControlAction(0, 135, true)  -- Control OVerride (Sub)
+            DisableControlAction(0, 200, true)  -- Pause Menu
+            DisableControlAction(0, 245, true)  -- Chat
+        else
+            Wait(100)
+        end
+    end
+end)
+
+RegisterNetEvent('npwd:disableControlActions')
```
itschip, line 41: We don't need to register it as a net event.

RockySouthpaw: Is it just an event?

itschip: Its client-to-client, so no need for it.

itschip:
> # Known Issues
> * If a server has any menu bound to any key that is typically used when writing a message (A-Z, commas, periods, ect) then that menu will pop up. This is a decent issue if you have `vMenu` or a similiar menu bound to one of these as closing it will also close the phone essentially making typing impossible. Majority of servers use M for `vMenu.`
> * People will move slightly every time they type the letters `W,A,S,D` as we do not freeze them. This just looks unnatural as they would end up doing 360 degree turns every sentence while also stepping into traffic. You would essentially only be able to text if you're in a corner or in the passenger seat.
> 
> These are the only things I've come across so far. I disabled most the controls I could think of while testing.
> 
> The only solution I have would be to disable all control actions if they're clicking on a text field or `SetNuiFocusKeepInput(false)` within text boxes.

Too bad

RockySouthpaw:
> > # Known Issues
> > 
> > * If a server has any menu bound to any key that is typically used when writing a message (A-Z, commas, periods, ect) then that menu will pop up. This is a decent issue if you have `vMenu` or a similiar menu bound to one of these as closing it will also close the phone essentially making typing impossible. Majority of servers use M for `vMenu.`
> > * People will move slightly every time they type the letters `W,A,S,D` as we do not freeze them. This just looks unnatural as they would end up doing 360 degree turns every sentence while also stepping into traffic. You would essentially only be able to text if you're in a corner or in the passenger seat.
> > 
> > These are the only things I've come across so far. I disabled most the controls I could think of while testing.
> > The only solution I have would be to disable all control actions if they're clicking on a text field or `SetNuiFocusKeepInput(false)` within text boxes.
> 
> Too bad

KEK

PR #232, opened 2021-05-20
feat(client/call): make pma-voice the default export

AvarianKnight:
mumble-voip is no longer maintained and is recommended to not be used as it is broken & doesn't have modern features.


RockySouthpaw:
I’ll leave this to @TasoOneAsia but this seems fine to me to change. It seems pretty stupid to support essentially a depreciated resource. Especially given that the forum page tells people to use pma instead. As seen [here](https://forum.cfx.re/t/release-mumble-voip-rp-radio/1083683).

We could always leave it and do a Config for it, but it’s relatively pointless to support it just out of a curtesy. We all already recommend PMA to most people who use mumble-voip to begin with.

itschip (approved):

PR #242, opened 2021-05-22
Bugfixes

AvarianKnight:


itschip:
Add some description to what this PR fixes.

jfrader (approved):

RockySouthpaw:
@AvarianKnight please list what this PR fixes.

AvarianKnight:
> @AvarianKnight please list what this PR fixes.

The commits have what each one fixes

RockySouthpaw:
> > @AvarianKnight please list what this PR fixes.
> 
> The commits have what each one fixes

Descriptions exist for this

resources/server/messages/messages.service.ts
```diff
@@ -120,17 +120,18 @@ class _MessagesService {
 
       for (const participantId of userParticipants) {
         // we don't broadcast to the source of the event.
-        if (participantId !== _identifier) {
-          const participantPlayer = PlayerService.getPlayerFromIdentifier(participantId);
-          if (!participantPlayer) {
```
jfrader, line 125: If you remove this return, whenever you send a message to an offline player you will get the catch statement, which is not intended in this case because its not an error, it's just the broadcast and the refresh that won't be triggered for the offline receiver.

jfrader (changes requested):
Except from one thing, everything else looks good.

What I mean is that your changes to send message don't handle offline players gracefully.

resources/server/messages/messages.service.ts
jfrader: Well, as talked offline, not only don't remove the return, but also changing it to a "continue" should do the trick.

TasoOneAsia:
@AvarianKnight expect mighty conflicts with this PR over the coming days

itschip:
@AvarianKnight Is the comment from @jfrader resolved?

PR #243, opened 2021-05-22
fix(phone/nav): prevent default so we don't break CEF

AvarianKnight:
Moving this to a separate branch as it may not be needed in the future, mainly want to leave it open for anyone wanting to currently test/use


lukealford:
I've tested this and can confirm this fixes CEF issues on shift-clicking.

No negatives or bugs I've found..

jfrader (approved):

TasoOneAsia:
This will be closed pending the merger of the FiveM PR 🤞

TasoOneAsia:
Fixed upstream with https://github.com/citizenfx/fivem/commit/34104fd5fc2d69caa6834d19f180498ce576888e

PR #254, opened 2021-06-04
Refactor/marketplace - Use promisfied net events and keep open HTTP req

TasoOneAsia:
Refactors to move towards more predictable and manageable flow on both the server and client. This uses promisified net events to callback to the original HTTP request directly, keeping it open. This increases ease of development quite a bit compared to the previous listener approach we took


itschip (approved):

itschip:
Just fix conflicts. Other than that, good!

PR #257, opened 2021-07-02
Offline Player Handling

### LINKED ISSUE #228: Bug(Calling): Offline players/invalid numbers are currently not handled when calling
### Developement Issue

Yes

### Phone Version

Develop

### Issue Description

When calling a player who is offline, the call fails and a server console error is shown.

### Issue Reproduction

- Open Dailer
- Call number who is offline

Receive error

### Evidence

image

image

### Suggested work around

Simply send a different error to the UI when source isn't found that the player is currently outside of the service area.

Refactor the console print stating player is offline.

AvarianKnight:
This is a pull request to add offline player handling, this isn't done and should be considered a draft.

Aims to fix #228

### DIFF
--- a/resources/server/players/player.service.ts
+++ b/resources/server/players/player.service.ts
@@ -63,8 +63,6 @@ class _PlayerService {
    */
   getPlayerFromIdentifier(identifier: string): Player | null {
     const player = this.playersByIdentifier.get(identifier);
-    if (!player)
-      throw new Error(`Could not find corresponding player for identifier: ${identifier}`);
     return player;
   }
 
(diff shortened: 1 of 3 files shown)


RockySouthpaw:
Tag me when this changes so I can remove the label :)

resources/server/players/player.service.ts
```diff
@@ -63,8 +63,6 @@ class _PlayerService {
    */
   getPlayerFromIdentifier(identifier: string): Player | null {
     const player = this.playersByIdentifier.get(identifier);
-    if (!player)
```
TasoOneAsia, line 66: Just return null if the player isn't found.

TasoOneAsia (changes requested):

AvarianKnight:
Closing this as this likely won't be needed with the new messaging refactor.

PR #258, opened 2021-07-02
Bug Fixes

### LINKED ISSUE #236: Bug: Multichar error
**Describe the bug**
There is an issue with getIdentifier being triggered, even if multichar support is enabled, it still try to call it and drops a error soup in console.

1621566925517c3df1fd39dc753cbd76cc5c105a71

**To Reproduce**
Using a system where it's not a normal identifier used.

### LINKED ISSUE #240: Enhancement: Better camera
The camera is very limited currently, you can only take one direction images, gcphones camera module, allows more freedome and ability to zoom, i think this should be looked in to, since this feels like a must have feature, since it makes the phone camera way more versatile.

### LINKED ISSUE #251: Bug: Tweets are sometimes initalized as null
**Describe the bug**
Tweets are sometimes null, which crash the entire phone

**To Reproduce**
** This isn't reliable at all **
1. First load in go to tweets
2. Tweets are null
3. Phone crashes

**Expected behavior**
Tweets to not be null on first initialization 

**Needed information (please complete the following information):**
 - **Server OS**: Ubuntu 20.04
- **Server Artifact version**: [e.g. 3486] 4000
- What version of **new-phone-who-dis** are you running? N/A

AvarianKnight:
Fixes some typings for calls

Fixes #236
Fixes #251
Fixes #240


itschip:
You mentioned previously that this fixes issues for multichar. Have you tested this on a server not using it?

AvarianKnight:
> You mentioned previously that this fixes issues for multichar. Have you tested this on a server not using it? (Aiming towards 'player ready' stuff)

No, I assumed (incorrectly from looking at the code) that something else handled player initilization

typings/call.ts
```diff
@@ -13,9 +13,9 @@ export interface CallHistoryItem {
   transmitterSource?: number;
   receiver: string;
   receiverSource?: number;
-  start: string;
+  start: string|number;
```
itschip, line 16: `start` is a string as we save it as a VARCHAR for the time being. Either parseInt it somwhere else in the client/server or just keep it as a number, as we parse it in the `DailerHistory` component.
Ref: https://github.com/project-error/npwd/blob/master/phone/src/apps/dialer/components/views/DialerHistory.tsx#L95

itschip (changes requested):

typings/call.ts
AvarianKnight: This was something I already talked to Taso about and he said this was the simplest solution, I can change it to a string that's what y'all want to do.

resources/server/players/player.controller.ts
```diff
@@ -47,7 +47,7 @@ if (!config.general.enableMultiChar) {
     if (resource === GetCurrentResourceName()) {
       const onlinePlayers = getPlayers();
       for (const player of onlinePlayers) {
-        await PlayerService.handleNewPlayerJoined(parseInt(player));
+        await PlayerService.handleNewPlayerJoined(player);
```
itschip, line 50: `getPlayers` still returns a array of strings. Is this done because you have updated types on your end?

AvarianKnight: don't know how this managed to sneak in

PR #259, opened 2021-07-06
feat(): add animation and call service & controllers

### LINKED ISSUE #249: BUG: Animations Reset
**Describe the bug**
Animations reset when doing any action

**To Reproduce**
1. Start a phone call
2. Close the phone
3. Animation goes away with the phone

**Expected behavior**
Animations to continue to play when in a phone call

AvarianKnight:
- run animations on interval so we don't lose animations when walking through doors
- move cl_call.ts to a class

This code still needs to be improved vastly, please leave comments on how it can be improved

This fixes #249


resources/client/cl_main.ts
```diff
@@ -1,13 +1,13 @@
 import { sendMessage } from '../utils/messages';
 import { PhoneEvents } from '../../typings/phone';
 import { TwitterEvents } from '../../typings/twitter';
-import { phoneCloseAnim, phoneOpenAnim, removePhoneProp } from './functions';
+import { removePhoneProp } from './functions';
```
itschip, line 4: Why is this imported here? I only see its usage in `animation.service.ts`

resources/client/animations/animation.service.ts
```diff
@@ -0,0 +1,194 @@
...
+      const playerPed = PlayerPedId();
+      if (this.onCall) {
+        this.handleCallAnimation(playerPed);
+      } else if (this.phoneOpen && !this.onCamera) {
+        this.handleOpenAnimation(playerPed);
+      }
+    }, 250);
+  }
+
+  setPhoneState(state: AnimationState, stateValue: boolean) {
```
itschip, line 27: Functions like these can be set to private as they should only be accessible within the class.

resources/client/animations/animation.service.ts
```diff
@@ -0,0 +1,194 @@
+import { newPhoneProp, removePhoneProp } from '../functions';
+import { Delay } from '../../utils/fivem';
+
+export enum AnimationState {
```
itschip, line 4: Maybe add these to `typings` for consistency?

resources/client/cl_main.ts
AvarianKnight: This was still here because it was used onResourceStop, though its still better to just convert it to use the animation service

resources/client/animations/animation.service.ts
AvarianKnight: The vast majority of this can be changed to be private, the only things that should be exposed is open/closePhone, and start/endPhoneCall

resources/client/animations/animation.service.ts
AvarianKnight: I left it out of typings mainly due to the fact that everything in typings is client & server, while this is only client.

Plus a file for a single enum isn't needed (IMO)

itschip (approved):

PR #287, opened 2021-07-23
Rebrand Iodine links to Fileglass

goodnightnoodles:
Iodine (iodine.gg) has rebranded to Fileglass (file.glass). Whilst the current Iodine links still work, they add extra, unnecessary overhead having to redirect from Iodine.


TasoOneAsia:
How long are those old links gonna be active for even. Just curious. Also odd commit convention choice.

goodnightnoodles:
> How long are those old links gonna be active for even. Just curious. Also odd commit convention choice.

Permanently until the domain expires early-mid 2022.

PR #295, opened 2021-08-03
Fix String parameter on getLocationFromUrl in order to be able to build.

YisusOnDev:
Error Argument of type 'String' is not assignable to parameter of type 'string'.
  'string' is a primitive, but 'String' is a wrapper object. Prefer using 'string' when possible.

Fix


TasoOneAsia:
This has already been fixed in the develop branch. This is a non-issue.

PR #317, opened 2021-09-24
Feat: Phone Number Database Config

RockySouthpaw:
- Adds a config option for the phone number column name within the config,json.

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,6 +13,7 @@
     "useIdentifierPrefix": false,
     "playerTable": "users",
     "identifierColumn": "identifier",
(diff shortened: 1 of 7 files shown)


TasoOneAsia (changes requested):
Make this a debug log not an info

TasoOneAsia:
Since GH mobile is a pain my review is targeting the latest commit

RockySouthpaw:
Removed the debug print. Someone will have to test it though as idk if the errors i get is due to dev branch being goofed or what:
image
image

PR #321, opened 2021-10-02
feat(client/photo): add the ability to use a custom image server

Ekinoxx0:
Credits to AvarianKnight for the base of this commit from [PR 256](https://github.com/project-error/npwd/pull/256)

This will allow changing upload url, field type for screenshot-basic (imgur or files[] for example)

Custom imageEncoding (Not sure why Avarian changed it to jpg by default as this is also the default in screenshot-basic as of this line : https://github.com/citizenfx/screenshot-basic/blob/0712d9ed5a0c6899db9bab37cd94f03310bb416c/src/client/client.ts#L51)

About returnedUrl**
Depending on the host, this could vary a lot, as we can't specify them all, I adopted the most known one
Imgur return a json with "data" object containing "link" as a field of this object.
Avarian selected host (we don't know the name) return a json with "files" array, the first object of this array containing the "url" field.
Discord webhooks also agree to this format, but is naming their array "attachments", the first obj of this array containing "proxy_url" field.

I think this variation can allow be agreed upon with this type of configuration.
There could about be variation in the authorization prefix so I made it configurable. (Bearer or API-ID are known prefix)

This is a long commit, with a lot a configurability in it, so it will probably need some testing before merging.

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,6 +17,17 @@
     "profileQueries": true,
     "phoneNumberColumn": "phone_number"
   },
+  "images": {
+    "url": "https://api.imgur.com/3/image",
+    "type": "imgur",
+    "imageEncoding": "jpg",
+    "contentType": "multipart/form-data",
+    "authorizationPrefix": "Client-ID",
+    "useAuthorization": true,
+    "returnedUrlObjectName": "data",
+    "returnedUrlFieldName": "link",
+    "returnedUrlInArray": false
+  },
   "twitter": {
     "showNotifications": true,
     "generateProfileNameFromUsers": true,
--- a/resources/client/cl_photo.ts
+++ b/resources/client/cl_photo.ts
@@ -2,7 +2,7 @@ import { PhotoEvents } from '../../typings/photo';
 import { Delay } from '../utils/fivem';
 import { sendCameraEvent, sendMessage } from '../utils/messages';
 import { PhoneEvents } from '../../typings/phone';
-import { ClUtils } from './client';
+import { ClUtils, config } from './client';
 import { animationService } from './animations/animation.controller';
 import { RegisterNuiCB, RegisterNuiProxy } from './cl_utils';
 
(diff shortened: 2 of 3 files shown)


TasoOneAsia:
I really want to unit test this but considering certain state of coverage, meh. @Ekinoxx0, can you list the test cases you have tested this against?

Ekinoxx0:
Hi (just trying to answer this quickly)

Tested this against :
 - Imgur (Working)
 - Discord webhook (Working)
 - ImageShack (Not working) Needs result.images[0].direct_link

This is a "good" first attempt but would need a different option style to parse all different types that could exist..
Also, we could use ShareX as a template to inspire and test this part of upload service :
https://github.com/ShareX/ShareX/tree/master/ShareX.UploadersLib/ImageUploaders

RockySouthpaw:
> Yeah I am concerned that we are adding something as a config option that is inherently unconfigurable easily, but not sure how to solve that to increase accessibility. Any thoughts?

I think this PR is the best attempt we’d get at that. There’s not much more that we can abstract to Config but this seems like something we certainly could. It would also mean that if the end user did adjust it, they wouldn’t have to rebuild which is the goal here too. In the event imgur goes down we should also have a way to readily adjust for the end user.

We can add the Config and update the docs with some example code to reflect the various verified config adjustments that would work while also denoting what would require additional code changes.

TasoOneAsia:
> In the event imgur goes down we should also have a way to readily adjust for the end user.

I agree, but this looks a little complicated to the end user configuring this resource, we can use it for now but should be simplified and expanded more in the future.

RockySouthpaw:
> > In the event imgur goes down we should also have a way to readily adjust for the end user.
> 
> I agree, but this looks a little complicated to the end user configuring this resource, we can use it for now but should be simplified and expanded more in the future.

Only part I really see as complicated would be: 
```
"contentType": "multipart/form-data",
     "authorizationPrefix": "Client-ID",
     "useAuthorization": true,
     "returnedUrlObjectName": "data",
     "returnedUrlFieldName": "link",
     "returnedUrlInArray": false
```
As messing with this seems a bit outside the scope of the average end user but idk exactly how much of this is even required to be changed depending on what image server to be used. (Discord, shareX ect)

Ekinoxx0:
Just made an other commit to make data parsing from upload services simpler.
This is also compatible with every upload services that return json content and is inspired from the ShareX examples as stated in my precedent message.

Using returnedDataIndexes as a route to parse, you can make it compatible with discord & others services like this :
`
    "returnedDataIndexes": [
      "attachments",
      0,
      "proxy_url"
    ]
`
Less variable and more flexible (the use of "any" type is needed for numbered indexes) 👍
Also a lot more readable

Let me know if you want this squashed in my first commit ;)

PR #322, opened 2021-10-03
fix: Add dependency into manifest

RockySouthpaw:
screenshot-basic is required for camera functionality so its a dependency and should be listed as such imo. I don't really see it being an optional thing since photos are going to be integrated into messages and at some point I think marketplace if I'm not mistaken.

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -19,3 +19,7 @@ files {
     'resources/html/index.html',
     'resources/html/**/*',
 }
+
+dependency {
+	'screenshot-basic'
+}


Ekinoxx0 (approved):
Seems legit :)
It could also pose questions about the dependence on ESX imposed by the bank controller (server side)

See : https://github.com/project-error/npwd/blob/2f168c3255a41a28b760b043ad35d8af5927990a/resources/server/bank/bank.controller.ts
And
https://github.com/project-error/npwd/blob/2f168c3255a41a28b760b043ad35d8af5927990a/resources/server/server.ts#L43

itschip:
> Seems legit :) It could also pose questions about the dependence on ESX imposed by the bank controller (server side)
> 
> See : https://github.com/project-error/npwd/blob/2f168c3255a41a28b760b043ad35d8af5927990a/resources/server/bank/bank.controller.ts And
> 
> https://github.com/project-error/npwd/blob/2f168c3255a41a28b760b043ad35d8af5927990a/resources/server/server.ts#L43

No, the bank controller is just dead, and has to be completely re-written. We're getting rid of ESX completely.

itschip (approved):

RockySouthpaw:
> > Seems legit :) It could also pose questions about the dependence on ESX imposed by the bank controller (server side)
> > See : https://github.com/project-error/npwd/blob/2f168c3255a41a28b760b043ad35d8af5927990a/resources/server/bank/bank.controller.ts And
> > https://github.com/project-error/npwd/blob/2f168c3255a41a28b760b043ad35d8af5927990a/resources/server/server.ts#L43
> 
> No, the bank controller is just dead, and has to be completely re-written. We're getting rid of ESX completely.

It also wouldn’t be really that difficult to as we can wrap certain things around configs so should be relatively straight forward to achieve once we get time to

itschip:
> It also wouldn’t be really that difficult to as we can wrap certain things around configs so should be relatively straight forward to achieve once we get time to

Might just end up writing an API for banking tbh.

TasoOneAsia:
ESX dependencies for bank are gonna be written out.

Issue #330, opened 2021-10-13
Bug: Eager NUI Requests are Invalid on Mount


TasoOneAsia:
### Developement Issue

No

### Issue Description

Fetching on the client side when using an async selector immediately occurs whenever a component is mounted that requires the scoped data. This creates a problem of eager fetching, as certain apps may immediately trigger a server side fetch upon NUI load before the player has been added to the server maps.

To fix this, we must prevent requests from going out until we can confirm that the player source has been processed and exists within the server side player maps. At this point, it is considered safe to initialize server requests.

### Issue Reproduction

N/A

RockySouthpaw:
Tracked here: https://app.gitkraken.com/glo/card/de59f3fd72894673bc953d3eb662f4f9

TasoOneAsia:
Closed with #334

PR #334, opened 2021-10-28
fix: Multicharacter loading and number generation

thelindat:
An event is triggered on the client after the player is added or removed, updating the `playerLoaded` value.
Default state of playerLoaded is false and, if not using multichar, will be changed by the playerSpawned event; not sure if it's necessary (not sure if #330 applies to regular spawning).

I'm not sure how `generateUniquePhoneNumber()` worked at all, but apparently it does for non-multichar? Didn't await result and just ended up querying with a promise as the first parameter; regardless, I needed to use rawExec to successfully query the database or I would get some error related to an undefined value and reading 'pool'. 😕

Refactor and improve this however since it could 100% be done cleaner and I'm not sure where you would want the playerReady function defined.

### DIFF
--- a/resources/client/cl_main.ts
+++ b/resources/client/cl_main.ts
@@ -180,3 +180,25 @@ RegisterNuiCB<{ keepGameFocus: boolean }>(
 //   const time = getCurrentGameTime()
 //   sendMessage('PHONE', 'setTime', time)
 // }, 2000);
+
+// Set a global function to check if the player is loaded
+let playerLoaded = false;
+(global as any).playerReady = async () => {
+  if (playerLoaded) return true;
+  return new Promise<boolean>((resolve) => {
+    setInterval(() => {
+      if (playerLoaded) resolve(true);
+    }, 50);
+  });
+};
+
+// Register events to set the state of playerLoaded
+if (!config.general.enableMultiChar) {
+  on(`playerSpawned`, async () => {
+    playerLoaded = true;
+  });
+} else {
+  onNet(PhoneEvents.PLAYER_LOADED, async (state: boolean) => {
+    playerLoaded = state;
+  });
+}
--- a/resources/client/cl_utils.ts
+++ b/resources/client/cl_utils.ts
@@ -1,3 +1,4 @@
+import { PhoneEvents } from '../../typings/phone';
 import { uuidv4 } from '../utils/fivem';
 import { ClUtils, config } from './client';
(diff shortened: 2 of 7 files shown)


itschip:
> I'll try out the qb wrapper later, think there may be an issue with receiving the phone number due to the function not expecting it handleNewPlayerEvent({ source: src, identifier, firstname, lastname }.

Shouldn't be an issue tho, we call the function with tries to find or generate a number.

thelindat:
> 
> 
> > I'll try out the qb wrapper later, think there may be an issue with receiving the phone number due to the function not expecting it handleNewPlayerEvent({ source: src, identifier, firstname, lastname }.
> 
> Shouldn't be an issue tho, we call the function with tries to find or generate a number.

For both QB and the ESX compatibility resources, they generate a number and send it to npwd (I did it to be in-line with qb); the reasoning being so we can set xPlayer.variables.phoneNumber / qbPlayer.PlayerData.charinfo.phone). It may be better to export the function for getting/generating a new number, however (we can export JS promises with artifacts from ~4500+).

resources/client/cl_main.ts
```diff
@@ -180,3 +180,25 @@ RegisterNuiCB<{ keepGameFocus: boolean }>(
...
+// Set a global function to check if the player is loaded
+let playerLoaded = false;
+(global as any).playerReady = async () => {
+  if (playerLoaded) return true;
+  return new Promise<boolean>((resolve) => {
+    setInterval(() => {
+      if (playerLoaded) resolve(true);
+    }, 50);
+  });
+};
```
TasoOneAsia, line 193: I meant lets store the flag globally. For example, `(global as any).playerLoaded`. Then in case we need to check the load state anywhere else we can just reference `(global as any).playerLoaded` as a flag.

TasoOneAsia (approved):
Even with that small comment about the global flag, this works as expected. We may need to be wary regarding the timeout of any proxied events as well.

itschip (approved):

resources/client/cl_main.ts
thelindat: The only issue is we have situations where awaiting `playerLoaded` may be necessary (proxy events being one, of course). To be honest it's a work-around, I believe it might be better to trigger a request to fetch from the client after the spawn event - but I assume that will require a bit of a rework on the NUI setup.

PR #335, opened 2021-11-01
feat(locale): add some french locale

Ekinoxx0:
Hi there,

I know this is absolutly not a priority, but I'm seeing this in my commit hist for some times now so here it is 👍 
French community ❤️ npwd

### DIFF
--- a/phone/src/locale/fr.json
+++ b/phone/src/locale/fr.json
@@ -9,6 +9,49 @@
         "INCOMING": "Appel entrant"
       }
     },
+    "NOTIFICATIONS": {
+      "NO_UNREAD": "Aucune notifications non lues"
+    },
+    "BROWSER": {
+      "NAME": "Navigateur"
+    },
+    "GENERIC": {
+      "LOADING": "Chargement...",
+      "WAIT": "Patientez...",
+      "SAVE": "Sauvegarder",
+      "EDIT": "Editer",
+      "UPDATE": "Mettre à jour",
+      "REQUIRED": "Requis",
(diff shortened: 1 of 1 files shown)


itschip:
Thanks for the PR, however this had to wait until we finish Twitter, so we can merge these in sync :D

itschip:
Uhm, not sure why these four last commits are even related to 'translation' lol

resources/server/utils/generateProfileName.ts
```diff
@@ -16,7 +16,7 @@ export async function generateProfileName(
   const phone_number = clean(player.getPhoneNumber());
 
   if (firstname && lastname) {
-    return `${firstname}${delimiter}${lastname}`;
+    return `${firstname}${delimiter}${lastname}${identifier}`;
```
itschip, line 19: Why do we want the identifier here?

Ekinoxx0:
My bad, I pushed commits without even thinking of this PR
Sorry for this, will fix this I soon as I can :)

resources/server/utils/generateProfileName.ts
Ekinoxx0: Even if this is not related to this PR (this is now deleted), just to answer : 
I'm currently using the multichar system as a multi simcard system, each "npwd character" is a new simcard in my framework, so identifier is the simcardId in my system. Adding identifier in this profile will fix a problem I have, if you have a character with multiple simcard, they will have the same profile name by default and this will lead to errors.

PR #337, opened 2021-11-03
Fix: contact querry based only off idenitfier

RockySouthpaw:
Resolves https://github.com/project-error/npwd/issues/336

 `number` within the sql should probably be changed to `contact_number` but this would mean some additional changes to accommodate this. Like adjusting [this](https://github.com/project-error/npwd/blob/a33f244a547559db7363d7388b1a05ad4cadbc02/resources/server/messages/messages.db.ts#L47) querry within messages and some other references. 
 
 Outside my expertise so I left it as number 🤷🏻.
 
 Edit: Theres also the issue of the code style within `contacts.services.ts` being non uniform so it's a bit weird at first. It needs some refactoring as well.

### DIFF
--- a/resources/server/contacts/contacts.service.ts
+++ b/resources/server/contacts/contacts.service.ts
@@ -17,8 +17,9 @@ class _ContactService {
     resp: PromiseEventResp<void>,
   ): Promise<void> {
     const identifier = PlayerService.getIdentifier(reqObj.source);
+    const srcPlayerNumber = PlayerService.getPhoneNumber(reqObj.source);
     try {
-      await this.contactsDB.updateContact(reqObj.data, identifier);
+      await this.contactsDB.updateContact(reqObj.data, identifier, srcPlayerNumber);
       resp({ status: 'ok' });
     } catch (e) {
       contactsLogger.error(`Error in handleUpdateContact (${identifier}), ${e.message}`);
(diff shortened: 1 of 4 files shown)


RockySouthpaw:
A similar fix also needs to be implemented for: 

- Twitter
- Match maker (I think?)
- Notes
- Messages
- Calls
- Gallery

essentially anywhere where the only identifier is license or whatever is set.

resources/server/contacts/contacts.service.ts
```diff
@@ -57,8 +60,10 @@ class _ContactService {
     resp: PromiseEventResp<Contact[]>,
   ): Promise<void> {
     const identifier = PlayerService.getIdentifier(reqObj.source);
+    const srcPlayerNumber = PlayerService.getPhoneNumber(reqObj.source);
+    console.log(srcPlayerNumber);
```
itschip, line 64: To be removed

itschip (changes requested):

resources/server/players/player.service.ts
```diff
@@ -57,6 +57,14 @@ class _PlayerService {
     return this.getPlayer(source).getIdentifier();
   }
 
+  /**
+   * Returns the player phone number for a given source
+   * Will return null if no player is found online with that source
+   **/
+  getPhoneNumber(source: number): string {
```
itschip, line 64: Do we really need this helper function? We already get it from the getPlayer metohd, and its feels like another layer of abstraction.

RockySouthpaw: Nah it’s not, will fix

resources/server/contacts/contacts.service.ts
RockySouthpaw: Resolved

itschip (approved):

PR #372, opened 2021-11-23
Tweak PLAYER_READY control flow & cleanup

thelindat:
Minor cleanup for loading playerdata (mostly related to multicharacter).

### DIFF
--- a/resources/client/cl_main.ts
+++ b/resources/client/cl_main.ts
@@ -1,7 +1,6 @@
 import { sendMessage } from '../utils/messages';
 import { PhoneEvents } from '../../typings/phone';
 import { TwitterEvents } from '../../typings/twitter';
-import { MessageEvents } from '../../typings/messages';
 import { config } from './client';
 import { animationService } from './animations/animation.controller';
 import { RegisterNuiCB } from './cl_utils';
@@ -9,9 +8,7 @@ import { RegisterNuiCB } from './cl_utils';
 // All main globals that are set and used across files
 (global as any).isPhoneOpen = false;
 (global as any).isPhoneDisabled = false;
-
-// Set a global function to check if the player has been loaded
-(global as any).isPlayerLoaded = config.general.enableMultiChar ? false : true;
+(global as any).isPlayerLoaded = false;
 
 const exps = global.exports;
 
--- a/resources/client/cl_main.ts
+++ b/resources/client/cl_main.ts
@@ -6,9 +6,9 @@ import { animationService } from './animations/animation.controller';
 import { RegisterNuiCB } from './cl_utils';
 
 // All main globals that are set and used across files
-(global as any).isPhoneOpen = false;
-(global as any).isPhoneDisabled = false;
-(global as any).isPlayerLoaded = false;
+global.isPhoneOpen = false;
+global.isPhoneDisabled = false;
+global.isPlayerLoaded = false;
 
 const exps = global.exports;
 
@@ -53,7 +53,7 @@ if (config.general.enableMultiChar) {
  * * * * * * * * * * * * */
 
 export const showPhone = async (): Promise<void> => {
-  (global as any).isPhoneOpen = true;
+  global.isPhoneOpen = true;
   const time = getCurrentGameTime();
   await animationService.openPhone(); // Animation starts before the phone is open
   emitNet(PhoneEvents.FETCH_CREDENTIALS);
@@ -66,7 +66,7 @@ export const showPhone = async (): Promise<void> => {
 };
 
 export const hidePhone = async (): Promise<void> => {
-  (global as any).isPhoneOpen = false;
+  global.isPhoneOpen = false;
   sendMessage('PHONE', PhoneEvents.SET_VISIBILITY, false);
   await animationService.closePhone();
   SetNuiFocus(false, false);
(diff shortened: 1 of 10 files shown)


resources/server/players/player.controller.ts
```diff
@@ -65,18 +65,18 @@ if (!config.general.enableMultiChar) {
 // instantiating/deleting a player. The config option must be set to true
 // for these to be available
 if (config.general.enableMultiChar) {
-  // This has to be an event as FXServer does not yet support exports which return a promise.
-  on('npwd:newPlayer', async (playerDTO: PlayerAddData) => {
+  const exp = global.exports;
```
TasoOneAsia, line 68: Lets move this alias into the global scope for consistency sake.

resources/server/players/player.controller.ts
```diff
@@ -65,18 +65,18 @@ if (!config.general.enableMultiChar) {
...
+
+  exp('newPlayer', async (playerDTO: PlayerAddData) => {
```
TasoOneAsia, line 70: Yay 🥳 I remember not doing these as exports because of our dear friend cfx_asyncRetVal. Appreciate the update.

resources/client/cl_main.ts
```diff
@@ -1,17 +1,14 @@
...
 import { animationService } from './animations/animation.controller';
 import { RegisterNuiCB } from './cl_utils';
 
 // All main globals that are set and used across files
 (global as any).isPhoneOpen = false;
 (global as any).isPhoneDisabled = false;
-
-// Set a global function to check if the player has been loaded
-(global as any).isPlayerLoaded = config.general.enableMultiChar ? false : true;
+(global as any).isPlayerLoaded = false;
```
TasoOneAsia, line 11: With the amount of flags we have now globally, lets actually be type safe and extend our global type to include our custom defined vars.

**global.d.ts** in `client/types/global.d.ts`

```ts
declare global {
  namespace NodeJS {
    interface Global {
      isPhoneOpen: boolean;
      isPhoneDisabled: boolean ;
    // ... etc
    }
  }
}
export {};
```

typings/phone.ts
```diff
@@ -13,7 +13,7 @@ export enum PhoneEvents {
   FETCH_CREDENTIALS = 'npwd:getCredentials',
   ON_INIT = 'npwd:onInit',
```
TasoOneAsia, line 14: Do we actually use the onInit event anywhere?

TasoOneAsia (changes requested):
Just a small suggestion for type safety of global assigned variables. Aside from that I think this looks good.

typings/phone.ts
thelindat: It's triggered by `handleNewPlayerJoined` for non-multicharacter, and `handleNewPlayerEvent` for multicharacter.

It triggers
```ts
function fetchOnInitialize() {
  emitNet(TwitterEvents.GET_OR_CREATE_PROFILE);
}
```

On that note, the following also triggers that function
```ts
RegisterCommand(
  'phone:restart',
  async () => {
    await hidePhone();
    sendMessage('PHONE', 'phoneRestart', {});
    fetchOnInitialize();
  },
  false,
);
```
but that sendMessage doesn't exist anywhere (I assume it was intended for resetting the UI?)

TasoOneAsia: This will most likely be killed with the twitter refactor and is legacy code from an older time. That onInit event is also from that time, which is why I ask.

```js
function fetchOnInitialize() {
  emitNet(TwitterEvents.GET_OR_CREATE_PROFILE);
}
```

TasoOneAsia: I prefer if we were to use PLAYER_LOADED and PLAYER_UNLOADED events or a combined SET_PLAYER_LOADED with param. onInit is weird naming from the past.

TasoOneAsia: @itschip Since twitter branch is under you, tagging you as this is relevant.

resources/server/players/player.service.ts
```diff
@@ -201,7 +201,7 @@ class _PlayerService {
    **/
   handleUnloadPlayerEvent(src: number) {
     this.deletePlayerFromMaps(src);
-
+    emitNet(PhoneEvents.PLAYER_UNLOADED, src, false);
```
itschip, line 204: Why are you passing an argument here? According to the `PLAYER_UNLOADED`  event, it does not have any parameters.

resources/client/cl_main.ts
```diff
@@ -25,9 +22,15 @@ function fetchOnInitialize() {
 }
 
 onNet(PhoneEvents.ON_INIT, () => {
+  (global as any).isPlayerLoaded = true;
+
   fetchOnInitialize();
 });
 
+onNet(PhoneEvents.PLAYER_UNLOADED, () => {
```
itschip, line 30: Should there be a param here?

itschip (changes requested):

resources/server/players/player.service.ts
thelindat: Leftover param from changing the event to be used when unloading, rather than toggling the state.

itschip: I believe we talked about aiming towards a removal of `ON_INIT`

TasoOneAsia (approved):

TasoOneAsia:
@RockySouthpaw if you are able to test these changes, I'll merge on your approval. If not, I'll merge when I get to chance to test it myself over the next couple days.

RockySouthpaw:
> @RockySouthpaw if you are able to test these changes, I'll merge on your approval. If not, I'll merge when I get to chance to test it myself over the next couple days.

Will test in about 4 and a half hours 👍🏻

Issue #364, opened 2021-11-22
Bug: Call notification isn't removed on rejection


RockySouthpaw:
### Developement Issue

No

### Issue Description

After rejecting a call, the person who rejected the call still has their notification for the call despite it having ended for both players and being removed from the map.

### Issue Reproduction

1. Player 1 calls Player 2
2. Player 2 rejects call

Call is ended for both players. Player 1 (person who called) doesn't have the notification alert while player 2 (the receiver) does.

### Evidence

![image](https://user-images.githubusercontent.com/55056068/142798927-a9d9b1ed-60e9-4d16-88f1-f5a34c1022d9.png)
![image](https://user-images.githubusercontent.com/55056068/142798944-ad129a11-61aa-4192-a324-890e7827014b.png)

TasoOneAsia:
Should be fixed in #374

PR #374, opened 2021-11-23
fix: emit CALL_ENDED and CALL_REJECTED events to both source and target

### LINKED ISSUE #332: Bug: Call doesn't end for both participants
### Developement Issue

Yes

### Issue Description

Calls doesnt working correctly on vrp base.

When player end the current call a promisse event npwd:endCall got status ok and user1 phone disconnected call, but the user2 continue call and when him click on endcall get error, the call doesnt exist anymore.

### Issue Reproduction

1. Login in 2 users
2. Make a call
3. End call
$. Who clicked will got success and disconnect call, but other contact doesnt update call status

### Evidence

unknown (2)

thelindat:
**Pull Request Description**
Should resolve #332 #364

**Pull Request Checklist**:

### DIFF
--- a/resources/server/calls/calls.service.ts
+++ b/resources/server/calls/calls.service.ts
@@ -197,6 +197,7 @@ class CallsService {
 
     // player who is calling and recieved the rejection.
     emitNet(CallEvents.WAS_REJECTED, currentCall.transmitterSource);
+    emitNet(CallEvents.WAS_REJECTED, currentCall.receiverSource);
 
     // Update our database
     await this.callsDB.updateCall(currentCall, false, endCallTimeUnix);
@@ -222,6 +223,7 @@ class CallsService {
     // lets protect against that
     if (currentCall?.is_accepted) {
       emitNet(CallEvents.WAS_ENDED, currentCall.receiverSource);
+      emitNet(CallEvents.WAS_ENDED, currentCall.transmitterSource);
     }
     // player who is calling (transmitter)
     resp({ status: 'ok' });


resources/server/calls/calls.service.ts
```diff
@@ -222,6 +223,7 @@ class CallsService {
     // lets protect against that
     if (currentCall?.is_accepted) {
       emitNet(CallEvents.WAS_ENDED, currentCall.receiverSource);
+      emitNet(CallEvents.WAS_ENDED, currentCall.transmitterSource);
     }
     // player who is calling (transmitter)
     resp({ status: 'ok' });
```
TasoOneAsia, line 229: Is there any point to this being a promiseMessageEvent if it isn't transactional? If we wish to go with classic events and their fire and forget behavior, then lets just make this a regular event handler.

TasoOneAsia (commented):
Just something to think about, but I believe this should fix the issues mentioned. Unable to test this till at least tomorrow though.

RockySouthpaw:
I’ll test this tonight and let you know @TasoOneAsia @thelindat

RockySouthpaw:
Does this resolve the issue where if a call is started and ended before the target player answers/rejects; the call doesn’t end for the target.

> Call logic probably needs to be refactored to where if the call state is still ringing then it needs to end the call for the receiver too.

TasoOneAsia:
The developer who implemented this part of the phone (shame on that Taso guy) should have explicitly defined all the possible states each participant can be in and made a nice map of  definitions but this wasn't done.

I'll see if I get all the possible states and outcomes mapped out, so we can be real *developers*

resources/server/calls/calls.service.ts
RockySouthpaw: I would assume 

> The developer who implemented this part of the phone (shame on that Taso guy) should have explicitly defined all the possible states each participant can be in and made a nice map of definitions but this wasn't done.
> 
> I'll see if I get all the possible states and outcomes mapped out, so we can be real _developers_

Would just be best if we had a simple util for `getPhoneState(playerId)`. We could probably use statebags for this actually I think. Not sure if that’s something you would want to add to the phone but we can do a state for inCall, and inPendingCall. And then just check the states of the player when someone is calling them and set it to whatever on the corresponding call events.

RockySouthpaw:
> The developer who implemented this part of the phone (shame on that Taso guy) should have explicitly defined all the possible states each participant can be in and made a nice map of definitions but this wasn't done.
> 
> I'll see if I get all the possible states and outcomes mapped out, so we can be real _developers_

Would just be best if we had a simple util for `getPhoneState(playerId)`. We could probably use statebags for this actually I think. Not sure if that’s something you would want to add to the phone but we can do a state for inCall, and inPendingCall. And then just check the states of the player when someone is calling them and set it to whatever on the corresponding call events.

RockySouthpaw:
> Does this resolve the issue where if a call is started and ended before the target player answers/rejects; the call doesn’t end for the target.
> 
> > Call logic probably needs to be refactored to where if the call state is still ringing then it needs to end the call for the receiver too.

it does not resolve this issue

RockySouthpaw (approved):

resources/server/calls/calls.service.ts
thelindat: I just threw in some events without looking at the rest of it 😆
Definitely need some changes.

thelindat:
> > Does this resolve the issue where if a call is started and ended before the target player answers/rejects; the call doesn’t end for the target.
> > > Call logic probably needs to be refactored to where if the call state is still ringing then it needs to end the call for the receiver too.
> 
> it does not resolve this issue

I had tested and rejected calls would properly end for both source and target but maybe I'm just 🤤

I'll go have a look.

thelindat:
> 
> I'll see if I get all the possible states and outcomes mapped out, so we can be real _developers_

it's 2021, real developers only release unfinished and buggy products.

RockySouthpaw:
> > > Does this resolve the issue where if a call is started and ended before the target player answers/rejects; the call doesn’t end for the target.
> > > > Call logic probably needs to be refactored to where if the call state is still ringing then it needs to end the call for the receiver too.
> > 
> > 
> > it does not resolve this issue
> 
> I had tested and rejected calls would properly end for both source and target but maybe I'm just 🤤
> 
> I'll go have a look.

They do end, however, I’m referring to this:
- Player 1 calls Player 2
- Player 2 doesn’t respond in any manner
- Player 1 ends call
- Player 2 will still have the call on their phone and they are unable to accept or deny it since it’s no longer in the call maps.

I’m sure other issues will arise later with this so it should also be addressed at some point but doesn’t need to be addressed in this PR unless you want to.

thelindat:
```ts
if (currentCall?.is_accepted) {
      emitNet(CallEvents.WAS_ENDED, currentCall.receiverSource);
      emitNet(CallEvents.WAS_ENDED, currentCall.transmitterSource);
    }
```

Changing this to `if (currentCall) {` seems to resolve it.

RockySouthpaw:
> ```ts
> ```ts
> if (currentCall?.is_accepted) {
>       emitNet(CallEvents.WAS_ENDED, currentCall.receiverSource);
>       emitNet(CallEvents.WAS_ENDED, currentCall.transmitterSource);
>     }
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> Changing this to if (currentCall) { seems to resolve it.
> ```

In testing I actually did that below the (!currentCall) check and it resolved it but It didn’t seem like the proper approach to me since my TS experience is limited and Im not sure if it would cause issues. But if it works in testing then LG2M.

resources/server/calls/calls.service.ts
itschip: If it 'definitely need some changes', lets make sure those are done, before merging. Rather fix it now, instead of having to go back and fix it again.

TasoOneAsia:
Confirmed working and ready for merge

PR #398, opened 2021-12-01
refactor: remove calls to DbInterface._rawExec

thelindat:
**Pull Request Description**

Start replacing the calls to the deprecated DbInterface._rawExec function and ensure proper typecasting.
- Utilise pool.execute over pool.query (may need to adjust some queries)
- Ensure proper return values/types

**Pull Request Checklist**:

### DIFF
--- a/resources/server/db/db_wrapper.ts
+++ b/resources/server/db/db_wrapper.ts
@@ -1,10 +1,7 @@
 import { pool } from './pool';
 import { mainLogger } from '../sv_logger';
-import { ResultSetHeader } from 'mysql2';
+import { ResultSetHeader, RowDataPacket } from 'mysql2';
 import { config } from '../server';
-import { withProfile } from '../utils/withProfile';
-
-const RESOURCE_NAME = GetCurrentResourceName();
 
 // Replicates ghmattimysql API
 class _DbInterface {
@@ -12,10 +9,19 @@ class _DbInterface {
 
   private async _internalQuery(query: string, values?: unknown) {
     try {
-      if (config.database.profileQueries) return await withProfile(pool.execute, query, values);
+      const conn = await pool.getConnection();
 
-      ScheduleResourceTick(RESOURCE_NAME);
-      return await pool.query(query, values);
+      if (config.database.profileQueries) {
+        const startTime = process.hrtime();
+        const result = await conn.execute(query, values);
+        this.logger.info(
+          `${query} (${values}) execution time: ${process.hrtime(startTime)[1] / 1000000}ms`,
+        );
+        return result;
+      }
+
+      conn.release();
+      return await pool.execute(query, values);
     } catch (e) {
       this.logger.error(`Error executing ${query} with error message ${e.message}`);
     }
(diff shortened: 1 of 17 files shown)


thelindat:
I'm testing all the queries at the moment, but I'm likely to miss stuff if @RockySouthpaw wants to double-check things.

TasoOneAsia:
Does latest node not require the scheduling tick to speed up IO callback dispatching?

thelindat:
> Does latest node not require the scheduling tick to speed up IO callback dispatching?

I don't believe it's necessary for internal use, rather when doing exports and having to wait for the response. Even without it in oxmysql it's fast until you throw a wait in your function.

resources/server/db/db_wrapper.ts
```diff
@@ -1,21 +1,28 @@
...
+        this.logger.info(
+          `${query} (${values}) execution time: ${process.hrtime(startTime)[1] / 1000000}ms`,
+        );
+        return result;
+      }
 
-      ScheduleResourceTick(RESOURCE_NAME);
-      return await pool.query(query, values);
+      const result = await conn.execute(query, values);
+      conn.release();
```
TasoOneAsia, line 24: Whats the reason for managing the connection release manually? Isn't this wrapped internally within `mysql` query wrapper?

resources/server/db/db_wrapper.ts
```diff
@@ -38,8 +44,8 @@ class _DbInterface {
    * @param values Variable definition
    **/
   public async exec(query: string, values?: unknown) {
-    const [res] = await this._internalQuery(query, values);
-    return (<ResultSetHeader>res).affectedRows;
+    const [res] = (await this._internalQuery(query, values)) as ResultSetHeader[];
+    return res.affectedRows;
   }
```
TasoOneAsia, line 49: The exec methods return types here could probably be done a little bit better here, using passed generics or something but tbh this works too, just explicit type assertion can feel a little *hacky* when compared to a generic system imo, need to think on long term implementation details but as a little hack this is fine.

resources/server/marketplace/marketplace.db.ts
```diff
@@ -23,44 +22,42 @@ export class _MarketplaceDB {
...
     const query = `DELETE FROM npwd_marketplace_listings WHERE identifier = ?`;
-    await DbInterface._rawExec(query, [identifier]);
+    await DbInterface.exec(query, [identifier]);
   }
 
   async getListing(listingId: number): Promise<MarketplaceListing> {
     const query = `SELECT * FROM npwd_marketplace_listings WHERE id = ?`;
-    const [results] = await DbInterface._rawExec(query, [listingId]);
+    const results = await DbInterface.fetch(query, [listingId]);
     const listings = <MarketplaceListing[]>results;
```
TasoOneAsia, line 49: As per last review comment, this method would benefit from a passable generic type here so we don't need to assert the result.

TasoOneAsia (commented):
Just a couple things to think about, do you have the updated IO cb query metrics when not manually scheduling another resource tick now?

thelindat:
Going to redo this stuff in the future once changes have settled down.

PR #403, opened 2021-12-03
feat(phone/photo): emit an event when exiting/entering photo mode

### LINKED ISSUE #384: Enhancement: Event for hiding ui components on picture taken
### Application

Camera

### Dependency

No

### Feature Description

Essentially a net event that is triggered when hud components should be hid. Like the `ClearHelp(bool);` function we have where the ui components are cleared before the "picture" is taken.

Would resolve issues such as peoples huds/maps being on pictures and would be super simple to add docs for. 

image

Would end up being something like 2 events. One for show and one for hide or we can just use a bool instead of 2 events.

Below is a lua example for how server owners could utilize it:
```lua
-- 2 Events
AddEventHandler('npwd:hideHudComponents', function ()
    DisplayRadar(false) -- Hides map
    -- Add hud here
end)

AddEventHandler('npwd:showHudComponents', function ()
    DisplayRadar(true) -- Shows map
     -- Add hud here
end)

-- 1 Event but use a bool for showing components
AddEventHandler('npwd:showHudComponents', function (visible)
    if visible then
        DisplayRadar(true) -- Show map
        -- show hud
    else
        DisplayRadar(false) -- hide map
        -- hide hud
    end
end)
```
This would be the best solution imo as server owners can fix their own problems without rebuilding phone.

TasoOneAsia:
**Pull Request Description**

Implements and closes #384 

**Pull Request Checklist**:

### DIFF
--- a/resources/client/cl_photo.ts
+++ b/resources/client/cl_photo.ts
@@ -43,10 +43,13 @@ RegisterNuiCB<void>(PhotoEvents.TAKE_PHOTO, async (_, cb) => {
   CellCamActivate(true, true);
   // Hide phone from rendering temporary
   closePhoneTemp();
-
   SetNuiFocus(false, false);
 
   inCameraMode = true;
+
+  // We want to emit this event for UI handling in other resources
+  // We hide nothing in NPWD by default
+  emit(PhotoEvents.NPWD_PHOTO_MODE_STARTED);
 
   while (inCameraMode) {
     await Delay(0);
@@ -68,10 +71,14 @@ RegisterNuiCB<void>(PhotoEvents.TAKE_PHOTO, async (_, cb) => {
     }
     displayHelperText();
   }
+
   ClearHelp(true);
+  // We can now signal to other resources for ending photo mode
+  // and redisplaying HUD components
+  emit(PhotoEvents.NPWD_PHOTO_MODE_ENDED);
+
   emit('npwd:disableControlActions', true);
   await animationService.closeCamera();
-  console.log('anim - closing camera');
 });
 
 const handleTakePicture = async () => {
--- a/typings/photo.ts
+++ b/typings/photo.ts
@@ -5,6 +5,8 @@ export interface GalleryPhoto {
 
 export enum PhotoEvents {
   TAKE_PHOTO = 'npwd:TakePhoto',
+  NPWD_PHOTO_MODE_STARTED = 'npwd:PhotoModeStarted',
+  NPWD_PHOTO_MODE_ENDED = 'npwd:PhotoModeEnded',
   TAKE_PHOTO_SUCCESS = 'npwd:TakePhotoSuccess',
   UPLOAD_PHOTO = 'npwd:UploadPhoto',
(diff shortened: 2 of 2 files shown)


resources/client/cl_photo.ts
```diff
@@ -43,11 +43,16 @@ RegisterNuiCB<void>(PhotoEvents.TAKE_PHOTO, async (_, cb) => {
...
-
   SetNuiFocus(false, false);
 
   inCameraMode = true;
 
+  // We want to emit this event for UI handling in other resources
+  emit(PhotoEvents.NPWD_PHOTO_MODE_STARTED);
+
+  // Player shouldnt have radar on screen
+  DisplayRadar(false);
```
RockySouthpaw, line 54: You shoukd check if the radar is already disabled before disabling it and then assign a bool here for later use if npwd disables it.

“why rocko?”
In the event a server has no minimap, we don’t want to enable it later. This would work as the initial check to see if they had a minimap to begin with. 

You can use `IsRadarHidden()` to check. It will return false if it’s showing.

resources/client/cl_photo.ts
```diff
@@ -68,10 +73,14 @@ RegisterNuiCB<void>(PhotoEvents.TAKE_PHOTO, async (_, cb) => {
     }
     displayHelperText();
   }
+
   ClearHelp(true);
+  // We can now signal to other resources for ending photo mode
+  emit(PhotoEvents.NPWD_PHOTO_MODE_ENDED);
+  DisplayRadar(true);
```
RockySouthpaw, line 80: Same here

RockySouthpaw (changes requested):

TasoOneAsia:
Actually thinking about this PR and your comments more, I feel that it would be best to leave this to third party scripts through just the event. Otherwise, no matter what, edge cases may appear regarding HUD management.

RockySouthpaw:
> Actually thinking about this PR and your comments more, I feel that it would be best to leave this to third party scripts through just the event. Otherwise, no matter what, edge cases may appear regarding HUD management.

Sg2m, will test shortly

RockySouthpaw:
Would it be bad to just pass a bool instead of using 2 events? 

Idk if you can help it but waiting on image uploading before showing hud components may not be the best idea imo as image uploading can get hung then the components will never show. 

Otherwise, this works well in testing. 

```lua
AddEventHandler('npwd:PhotoModeStarted', function()
    LocalPlayer.state:set('pictureMode', true, true)
    DisplayRadar(false)
    RemoveMultiplayerHudCash()
    RemoveMultiplayerBankCash()
    TriggerEvent('WH:hideHud')
end)

AddEventHandler('npwd:PhotoModeEnded', function ()
    LocalPlayer.state:set('pictureMode', false, true)
    if not Config.hideRadar then DisplayRadar(true) end
    TriggerEvent('WH:showHud')
end)
```

https://i.imgur.com/bxqTwIR.jpg

RockySouthpaw (approved):

PR #404, opened 2021-12-03
fix(server/marketplace): add else statement for reporting listing

RockySouthpaw:
Previously these would trigger regardless if a listing was reported or not. This resolves that issue

**Pull Request Checklist**:
  * [X] Have you followed the guidelines in our Contributing document and Code of Conduct?
* [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/project-error/npwd/pulls) for the same update/change?
* [X] Have you built and tested NPWD in-game after the relevant change?

### DIFF
--- a/resources/server/marketplace/marketplace.service.ts
+++ b/resources/server/marketplace/marketplace.service.ts
@@ -126,10 +126,10 @@ class _MarketplaceService {
       if (reportExists) {
         marketplaceLogger.error(`This listing has already been reported`);
         resp({ status: 'error', errorMsg: 'REPORT_EXISTS' });
+      } else {
+        await this.marketplaceDB.reportListing(rListing.id, rListing.name || rListing.username);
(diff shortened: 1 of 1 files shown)


resources/server/marketplace/marketplace.service.ts
```diff
@@ -126,10 +126,10 @@ class _MarketplaceService {
       if (reportExists) {
         marketplaceLogger.error(`This listing has already been reported`);
         resp({ status: 'error', errorMsg: 'REPORT_EXISTS' });
```
TasoOneAsia, line 128: Instead of adding that else statement we can also just add `return` keyword here.

itschip (approved):

resources/server/marketplace/marketplace.service.ts
RockySouthpaw: whatever you prefer 🤷🏻

PR #458, opened 2021-12-31
[Calling] Feat: Dialing Tone during Outgoing Request Phase

TasoOneAsia:
**Pull Request Description**

Adds a tone during the dialing phase of an outgoing request. This ends once the phone call is ended or once it is accepted.

**Pull Request Checklist**:

### DIFF
--- a/phone/src/os/call/hooks/useCall.ts
+++ b/phone/src/os/call/hooks/useCall.ts
@@ -2,11 +2,12 @@ import { ActiveCall } from '@typings/call';
 import { useCurrentCall } from './state';
 import { CallEvents } from '@typings/call';
 import { fetchNui } from '../../../utils/fetchNui';
-import { useCallback } from 'react';
+import { useCallback, useEffect } from 'react';
 import { useMyPhoneNumber } from '@os/simcard/hooks/useMyPhoneNumber';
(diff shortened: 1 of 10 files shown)


itschip (approved):

phone/src/os/call/hooks/useCall.ts
```diff
@@ -21,11 +22,18 @@ interface CallHook {
...
   const [call, setCall] = useCurrentCall();
-  // const [dialRing, setDialRing] = useState(false);
   const myPhoneNumber = useMyPhoneNumber();
   const [t] = useTranslation();
   const { addAlert } = useSnackbar();
-  // const { endDialTone, startDialTone } = useDialingSound();
+  const { endDialTone, startDialTone } = useDialingSound();
+
+  useEffect(() => {
+    if (call?.isTransmitter && !call.is_accepted) {
```
itschip, line 31: Can't `is_accepted` also be null here too?

TasoOneAsia: Yes but this conditional should handle for that, does it not?. I do think that our call state API needs a more declarative interface though?

itschip: Yes, the interface has to be improved. We can just keep it as it is. This is going to develop so, we can test there.

PR #461, opened 2022-01-01
fix(sql): Add column

RockySouthpaw:
**Pull Request Description**

**Pull Request Checklist**:
* [ ] Have you followed the guidelines in our Contributing document and Code of Conduct?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/project-error/npwd/pulls) for the same update/change?
* [ ] Have you built and tested NPWD in-game after the relevant change?

### DIFF
--- a/import.sql
+++ b/import.sql
@@ -1,6 +1,6 @@
 # comment this out if your server has a different table/location for phone_number
 # or if it already exists in the users table
-# ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) DEFAULT NULL;
+ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) DEFAULT NULL;
 
 CREATE TABLE IF NOT EXISTS `npwd_phone_contacts`
 (


RockySouthpaw:
Probably needs to be adjusted to check for the current table or we could remove this querry all together and handle this within starting the phone so the column exist based off the configured database stuff.

import.sql
```diff
@@ -1,6 +1,6 @@
 # comment this out if your server has a different table/location for phone_number
 # or if it already exists in the users table
-# ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) DEFAULT NULL;
+ALTER TABLE `users` ADD COLUMN `phone_number` VARCHAR(20) DEFAULT NULL;
```
itschip, line 3: We have to check if the table exists.

itschip (changes requested):

import.sql
RockySouthpaw: > We have to check if the table exists.

Isnt those checks dependent on MYSQL versions?

itschip: I don't think 'if exists' is

itschip: I don't think 'if exists' is

TasoOneAsia:
Something for me to tackle during stream today

RockySouthpaw:
> Something for me to tackle during stream today

And did you?

itschip:
> > Something for me to tackle during stream today
> 
> And did you?

smh

itschip:
@RockySouthpaw Just run a 'if exsits' and call it a day

RockySouthpaw:
effort

PR #477, opened 2022-01-04
feat: basic image filtering

AvarianKnight:
This is a basic PR to get general ideas to improve "image safety".

The main current issue is that users are allowed to post images from any website, this isn't something that's unique to npwd and is something a lot of larger tech companies have something in place for. See `https://images-ext-2.discordapp.net/external/kPbHbSF3VICIZj3y5aDagCzU891BqQcHBtMzz3ygEkE/https/i.dillonskaggs.dev/RpptWjd0?width=245&height=300`

Why is this an issue?

The most obvious thing is that bad actors can IP dump the entire server if they have their own image server setup.

Current Missing Features:

- [x] update to handle multiple Twitter images
- [x] Filter match makers profile images
- [x] Filter Twitters avatar profile images
- [x] Filter marketplace images
- [x] Filter Contact images
- [x] Filter Gallery images
- [x] Return better error messages for declined images

### DIFF
--- a/config.json
+++ b/config.json
@@ -26,6 +26,20 @@
     "useAuthorization": true,
     "returnedDataIndexes": ["data", "link"]
   },
+  "imageSafety": {
+    "filterUnsafeImageUrls": true,
+    "embedUnsafeImages": false,
+    "embedUrl": "https://i.example.com/embed",
+    "safeImageUrls": [
+      "imgur.com",
+      "file.glass",
+      "dropbox.com",
+      "tenor.com",
+      "discord.com",
+      "discordapp.com",
+      "wikipedia.org"
+    ]
+  },
   "twitter": {
     "showNotifications": true,
     "generateProfileNameFromUsers": true,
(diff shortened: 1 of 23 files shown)


RockySouthpaw:
You will also need to filter contact pictures as well

AvarianKnight:
> You will also need to filter contact pictures as well. Can also integrate discord logging into this too to notify of potential bad actors.

Don't think contacts get used by anything under than the local user

RockySouthpaw:
> > You will also need to filter contact pictures as well. Can also integrate discord logging into this too to notify of potential bad actors.
> 
> Don't think contacts get used by anything under than the local user

If we add the ability to share contacts, which I think is currently planned, it would no longer just be local user. Issue for a later date ofc but may be best to tackle it here anyways.

itschip:
@AvarianKnight Please rebase this, so we can moorge

itschip (approved):

Issue #426, opened 2021-12-12
Enhancement(Messages): Add a way to call from conversation


RockySouthpaw:
### Application

Messages

### Dependency

No

### Feature Description

Just add an option to call the player. We have the drop down functionality already so starting a call with the target number should be simple to add here. Could also do a button instead for calling and would give us enough padding

![image](https://user-images.githubusercontent.com/55056068/145701232-38f921c4-7dd8-417b-9884-72e9f37bd344.png)

NuttyShrimp:
Is something like this not more intuitive for the user?
![image](https://user-images.githubusercontent.com/5462901/147860712-4740beea-b6fd-4569-945f-70265d6ebb32.png)

RockySouthpaw:
The only reason I suggested using the existing drop down stuff is to not add anything additional functionality wise.

NuttyShrimp:
Something like this then? Only 'problem' I had to prevent adding a new translation was putting the phonenumber behind the translation
![image](https://user-images.githubusercontent.com/5462901/147966903-32727a6b-c9b1-48ce-9db6-102ef6e3ac64.png)

RockySouthpaw:
> Something like this then? Only 'problem' I had to prevent adding a new translation was putting the phonenumber behind the translation
> ![image](https://user-images.githubusercontent.com/5462901/147966903-32727a6b-c9b1-48ce-9db6-102ef6e3ac64.png)

Yea something like this would probably work okay. I think there might be some existing translations we can use for “call” I think. Not 100% sure though.

@TasoOneAsia @itschip Thoughts?

NuttyShrimp:
I'm currently using that translation but I don't now if there are languages where the number is placed somewhere else in the sentence then at the end

itschip:
> Is something like this not more intuitive for the user? ![image](https://user-images.githubusercontent.com/5462901/147860712-4740beea-b6fd-4569-945f-70265d6ebb32.png)

I like this. The modal should be deleted, as you can now delete multiple conversations from the messages list.

RockySouthpaw:
> > Is something like this not more intuitive for the user? ![image](https://user-images.githubusercontent.com/5462901/147860712-4740beea-b6fd-4569-945f-70265d6ebb32.png)
> 
> I like this. The modal should be deleted, as you can now delete multiple conversations from the messages list.
> 
> EDIT: Just make the icon slightly smaller.

[...]

PR #495, opened 2022-01-06
refactor(messages): remove modal & replace with icons

NuttyShrimp:
As discussed in #426 

**Pull Request Description**

**Pull Request Checklist**:

### DIFF
--- a/phone/src/apps/messages/components/modal/MessageModal.tsx
+++ b/phone/src/apps/messages/components/modal/MessageModal.tsx
@@ -22,6 +22,8 @@ import Modal from '../../../../ui/components/Modal';
 import { useMessagesState } from '../../hooks/state';
 import { makeStyles } from '@mui/styles';
 import { useMessageAPI } from '../../hooks/useMessageAPI';
+import { useCall } from '@os/call/hooks/useCall';
+import { Call } from '@mui/icons-material';
 
 const LARGE_HEADER_CHARS = 30;
 const MAX_HEADER_CHARS = 80;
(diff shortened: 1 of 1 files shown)


itschip:
Do you have an image on how this looks?

NuttyShrimp:
Yes. I used a smaller fontsize than I original showed in the issue because you mentioned it.
Only 'issue' I see is that the ripple around the button is not a perfect circle but I think that's more a MUI issue
image

itschip:
While we're one this, you can probably just remove the whole modal and the `MoreVert` icon stuff

NuttyShrimp:
So, I am currently replacing the modal but I can already see people accidentally clicking the wrong button when trying to call someone and removing there conversation.
So do I keep the modal and replace the contents to act like a confirm box to prevent users from accidentally removing there conversation?

itschip:
> So, I am currently replacing the modal but I can already see people accidentally clicking the wrong button when trying to call someone and removing there conversation. So do I keep the modal and replace the contents to act like a confirm box to prevent users from accidentally removing there conversation?

There won't be any way to 'accidentally clicking the wrong button' when there only will be one. Well, two if you don't have the person as a contact, but the group actions modal should not exist at all. Nor should the 'MoreVert' icon button either.

phone/src/apps/messages/components/modal/MessageModal.tsx
```diff
@@ -176,11 +151,11 @@ export const MessageModal = () => {
             </Typography>
             <Tooltip
               classes={{ tooltip: classes.tooltip }}
-              title={t('MESSAGES.ACTIONS_TITLE')}
-              placement="left"
+              title={`${t('DIALER.NAVBAR_DIAL')} ${targetNumber}`}
```
itschip, line 154: Let's use the generic translation key here :)

```suggestion
              title={`${t('GENERIC.CALL')} ${targetNumber}`}
```

itschip (changes requested):

phone/src/apps/messages/components/modal/MessageModal.tsx
NuttyShrimp: Ah I was searching for it but couldn't find it.

itschip (approved):

Issue #450, opened 2021-12-28
Bug(Marketplace): Gallery selection clears typed text.


RockySouthpaw:
### Developement Issue

Yes

### Phone Version

Develop

### Issue Description

When using marketplace and posting a listing, any typed text is gone when you add an image from gallery as it looks like the data is refreshed.

### Issue Reproduction

Type in a field and upload an image.

PR #496, opened 2022-01-06
fix(marketplace): fixed input field resetting after image selection

NuttyShrimp:
**Pull Request Description**
This is a fix for issue #450.
I can redo it with just React state instead of the recoil state but it would only complicate things imo.
It also clears the field after a succesful post was created

**Pull Request Checklist**:


AvarianKnight:
@NuttyShrimp I think this PR is named wrong?

NuttyShrimp:
Yep, my bad. Didn't double check before PRing

itschip (approved):
Tested and works :)

PR #524, opened 2022-01-11
Message Fixes

AvarianKnight:
**Pull Request Description**

This provides a fix for read state on messages, as previously they didn't work.

This also includes message ordering based on last activity

This is also a critical fix for group message creations, current `message_group_conversations` will have to be wiped due to a bug  that resulted in them not being created properly

**Pull Request Checklist**:


resources/server/messages/messages.db.ts
```diff
@@ -147,18 +146,30 @@ export class _MessagesDB {
...
     return result[0].count;
   }
-
+  /**
+   * Adds an unread to the specified group id
+   * @param groupId The unique group ID for the message
+   * @param phoneNumber The phone number of the player to ignore
+   */
+  async setMessageUnread(groupId: string, phoneNumber: string) {
+    console.log(groupId, phoneNumber);
```
itschip, line 155: Remove

resources/server/messages/messages.utils.ts
```diff
@@ -108,7 +109,7 @@ export async function createMessageGroupsFromPhoneNumber(
       sourcePhoneNumber,
     );
   } else {
-    await MessagesDB.createMessageGroup(sourcePhoneNumber, conversationId, sourcePhoneNumber);
+    await MessagesDB.createMessageGroup(tgtPhoneNumber, conversationId, sourcePhoneNumber);
```
itschip, line 112: With this change, have you tested cases like, if you send a message to a person who does not have a conversation with you? Does it create the group properly?

resources/server/messages/messages.service.ts
```diff
@@ -135,7 +143,7 @@ class _MessagesService {
           const participantPlayer = PlayerService.getPlayerFromIdentifier(participantIdentifier);
 
           if (participantPlayer) {
-            emitNet(MessageEvents.SEND_MESSAGE_SUCCESS, participantPlayer.source, messageData);
+            emitNet(MessageEvents.SEND_MESSAGE_SUCCESS, reqObj.source, messageData);
```
itschip, line 146: What's going on here? We still need to send this to the 'participant'. The broadcast doesn't update any messages.

itschip (changes requested):

resources/server/messages/messages.utils.ts
AvarianKnight: Yes it does

resources/server/messages/messages.db.ts
AvarianKnight: print go brrrrrrrrrrrr

PR #553, opened 2022-01-18
fix(call): fix being able to accept calls if player has no phone

AvarianKnight:
**Pull Request Description**
Title says it all :P

**Pull Request Checklist**:

### DIFF
--- a/resources/client/calls/cl_calls.service.ts
+++ b/resources/client/calls/cl_calls.service.ts
@@ -1,3 +1,4 @@
+import { checkHasPhone } from '../cl_main';
 import { IAlertProps } from '../../../typings/alerts';
 import { ActiveCall, CallEvents, CallRejectReasons } from '../../../typings/call';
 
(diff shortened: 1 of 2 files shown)


RockySouthpaw:
This seems weird to even handle. People should just set the phone disabled if the item is removed and we should check if the phone is disabled on emiting messages and calls. 

Super weird to do an item check every single time especially when this shitty frameworks use server callbacks for them.

AvarianKnight:
Both ways are not the greatest tbh, but this one is lighter on the server, at least for my framework as we don't use shitty callbacks to check items.

With setting the phone as disabled you then have the issue of people forgetting to change it back, so using the same export on the server would probably be a better idea.

RockySouthpaw:
> Both ways are not the greatest tbh, but this one is lighter on the server, at least for my framework as we don't use shitty callbacks to check items.
> 
> With setting the phone as disabled you then have the issue of people forgetting to change it back, so using the same export on the server would probably be a better idea.

idk we need something that people with these xd RP frameworks won’t experience potential performance issues on larger pops

RockySouthpaw:
@AvarianKnight Should the same logic be applied to notifications too? Or should that be dependent on phoneDisabled being true?

PR #555, opened 2022-01-20
feat(server/build-detector): inform user of missed build step

TasoOneAsia:
**Pull Request Description**

Explicitly convey to the user that using the release is a required step, literally giving them a direct DL link to the latest GH release.

**Pull Request Checklist**:

### DIFF
--- a/build-detector.js
+++ b/build-detector.js
@@ -1,0 +1,30 @@
+const fs = require('fs');
+const path = require('path');
+
+const currentResourceName = GetCurrentResourceName();
+
+// I hate that this file is necessary :(
+
+// This isn't included in release builds or in build process
+// as its only purpose is to detect whether we live in the non-built
+// source code.
+
+on('onResourceStart', async (resName) => {
+  if (resName !== currentResourceName && process.env.NODE_ENV !== 'development') return;
+
+  const currentResourcePath = GetResourcePath(currentResourceName);
+  const targetBuildPath = path.join(currentResourcePath, 'resources', 'dist');
+
+  try {
+    await fs.promises.access(targetBuildPath);
+  } catch {
+    console.log(
+      '^1==============================================\n\n' +
+        '!!! NPWD WAS UNABLE TO LOCATE BUILT FILES AND WAS UNABLE TO START !!!\n\n' +
+        'This error is most likely caused by downloading the source code instead of the release\n\n' +
+        'Find the latest NPWD release here: \n\n' +
+        'https://github.com/project-error/npwd/releases/latest/download/npwd.zip\n\n' +
+        '==============================================^0',
+    );
+  }
+});
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -9,6 +9,9 @@ client_scripts {
 }
 
 server_script {
+    -- This is a file that lives purely in source code and isn't compiled alongside
+    -- rest of the release. It's used to detect whether a user can read or not.
+    'build-detector.js',
     'resources/dist/server.js'
 }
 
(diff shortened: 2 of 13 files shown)


build-detector.js
```diff
@@ -0,0 +1,30 @@
...
+
+on('onResourceStart', async (resName) => {
+  if (resName !== currentResourceName && process.env.NODE_ENV !== 'development') return;
+
+  const currentResourcePath = GetResourcePath(currentResourceName);
+  const targetBuildPath = path.join(currentResourcePath, 'resources', 'dist');
+
+  try {
+    await fs.promises.access(targetBuildPath);
+  } catch {
```
RockySouthpaw, line 20: Should we also toss a confirmation message saying it contains built files or something letting them know they can read. Easier for debugging.

build-detector.js
```diff
@@ -0,0 +1,30 @@
...
+    console.log(
+      '^1==============================================\n\n' +
+        '!!! NPWD WAS UNABLE TO LOCATE BUILT FILES AND WAS UNABLE TO START !!!\n\n' +
+        'This error is most likely caused by downloading the source code instead of the release\n\n' +
```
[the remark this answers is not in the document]
TasoOneAsia: Can also arise due to error in build process.

build-detector.js
TasoOneAsia: Services are logged under the "info" level when they start.

build-detector.js
```diff
@@ -0,0 +1,30 @@
...
+        'Find the latest NPWD release here: \n\n' +
+        'https://github.com/project-error/npwd/releases/latest/download/npwd.zip\n\n' +
```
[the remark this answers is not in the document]
TasoOneAsia: Opening a browser tab during startup process might tick off server owners a good amount, could be something to consider but should be doing this in a way that does not mess with the host machine any.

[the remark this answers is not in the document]
RockySouthpaw: If only there was clipboard api

build-detector.js
RockySouthpaw: image

Yea but they're quite obnoxious and i've never recalled a time where these actually said they didn't start lol.

RockySouthpaw: Meh fuck it

RockySouthpaw:
There’s also a potential issue with this in terms of being misleading or server owners not seeing this print, though it’s not really a fault on the resource. 

The error message states NPWD failed to start however, it’s actually started. So any resource, like esx-npwd or custom frameworks, may encounter fatal issues if they’re relying on `if GetResourceState('npwd') == 'started' then` to let the resource know the resource was started as any sort of error with whatever pertains to the above check will result in the subsequent code not being executed.

This will also result in subsequent errors due to a lack of building like “no export newPlayer” which would also give us a bit of a harder time with support as people may be asking why framework integration isn’t working.

Ideally, we would stop the resource since it wasn’t built correctly to prevent any errors or misbehaving of reliant resources. 

If you’re thinking “how will a server owner not see a multi line print?” you haven’t seen some of the console logs these server owners have:

RockySouthpaw:
There’s also some cases of extremely obnoxious resources doing multi line prints, like essentialmode for instance, that just spells their names or whatever branding. We should also account for those people when implementing this so it may be worth it to run a thread if we can’t stop the resource altering the server owner of their issue.

TasoOneAsia:
This is as far as we can go in terms of handling for building. Anything more and it becomes tediously interdependent and will devolve into spaghetti over time.

RockySouthpaw:
Closed in favor of https://github.com/project-error/npwd/pull/559 due to me breaking this kek

PR #579, opened 2022-01-31
onMessage export

itschip:
Introduces an export for listening to messages, with limited features. This is due to the increasing request being able to, and I quote: "do roleplay stuff"


itschip:
Writing some quick docs for this as well.

itschip:
We also don't throw an error when the phone number doesn't exist.

PR #602, opened 2022-02-11
Debug stuff

c-wide:
**Pull Request Description**

This PR adds a boot controller that will perform various tasks/checks when the server resource starts and also adds a simple export check if PhoneAsItem is enabled.

This is WIP for a few reasons:

- Do you guys want this to be a controller/service or just a bunch of util functions I run in the server.ts file on startup?
- If it should be a controller/service, is "boot" the correct name? I think it accurately describes whats happening but I would like input.
- If boot is fine, should I remove the onServerResourceStart event handler from server.ts and provide confirmation after checks have been completed?
- After each check should I be providing confirmation that it passed or just log when an error occurs?
- How should the logs be formatted? Do we want a huge box like build-checker or just a simple colored log?
- Where should I put "requiredDbColumns" and "frameworkDependencies" that are located at the top of the boot service file or is this location fine?
- In cl_main I added an export checker if PhoneAsItem.enabled is true, should I make an event and also log the error to the server or is just client logging fine for this?

**Pull Request Checklist**:

### DIFF
--- a/resources/server/boot/boot.service.ts
+++ b/resources/server/boot/boot.service.ts
@@ -6,8 +6,8 @@ const { identifierColumn, phoneNumberColumn } = config.database;
 const requiredDbColumns = [identifierColumn, phoneNumberColumn];
 
 const frameworkDependencies = {
-  es_extended: ['esx-npwd'],
-  'qb-core': ['qb-npwd'],
+  ['es_extended']: ['esx-npwd'],
+  ['qb-core']: ['qb-npwd'],
 };
 
 export class _BootService {
--- a/resources/server/boot/boot.service.ts
+++ b/resources/server/boot/boot.service.ts
@@ -56,23 +56,21 @@ export class _BootService {
    * Check if various framework wrappers are started if applicable.
    */
   checkFrameworkDependencies(): void {
-    let startedResources: string[] = [];
+    const startedResources = new Set<string>();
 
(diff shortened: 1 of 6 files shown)


resources/server/boot/boot.service.ts
```diff
@@ -0,0 +1,86 @@
+import BootDb, { _BootDb } from './boot.db';
+import { bootLogger } from './boot.utils';
+import { config } from '../server';
+
+const { identifierColumn, phoneNumberColumn } = config.database;
+const requiredDbColumns = [identifierColumn, phoneNumberColumn];
+
+const frameworkDependencies = {
+  es_extended: ['esx-npwd'],
```
RockySouthpaw, line 9: Should this not be 'es_extended'

c-wide: The prettier config auto corrects it

RockySouthpaw: Well qb-core has '' so kind of weird es_extended wouldn’t

itschip: @RockySouthpaw It's because the `_` is valid in a object property, while `-` is not :/

c-wide:
Requested changes made

c-wide:
Should be good for review/testing and merge

itschip (approved):
Tested and seems to work fine. LGTM :)

PR #604, opened 2022-02-12
Match update

Patilhas:
Add match notification and fix update new accounts

**Pull Request Description**

**Pull Request Checklist**:

This change enables notifications in match to the other player. For example, player A has matched player B in the past, then if player B matches player A then player A receives a notification.

This also brings a fix that if a new match account is created after the other players have loaded the possible matches it wouldn't load to that players the new account.

### DIFF
--- a/phone/src/apps/match/hooks/useMatchActions.ts
+++ b/phone/src/apps/match/hooks/useMatchActions.ts
@@ -51,17 +51,21 @@ export const useMatchActions = () => {
     [setProfiles, addAlert, t, setMatches],
   );
 
-  const addMatchAccount = useRecoilCallback(({ snapshot, set }) => async (profile: FormattedProfile, myProfile: FormattedProfile) => {
-    if (profile.identifier === myProfile.identifier) return;
+  const addMatchAccount = useRecoilCallback(
+    ({ snapshot, set }) =>
+      async (profile: FormattedProfile, myProfile: FormattedProfile) => {
+        if (profile.identifier === myProfile.identifier) return;
 
-    const matchesLoading = getIsMatchesLoading(snapshot);
-    if (matchesLoading) return;
+        const matchesLoading = getIsMatchesLoading(snapshot);
+        if (matchesLoading) return;
 
-    set(matchState.profiles, (curVal) => [profile, ...curVal]);
-  });
+        set(matchState.profiles, (curVal) => [profile, ...curVal]);
+      },
+  );
 
   const addMatchedAccount = async () => {
     fetchNui<ServerPromiseResp<FormattedMatch[]>>(MatchEvents.GET_MATCHES).then((resp) => {
+      if (resp.status !== 'ok') return;
       setMatches(resp.data);
     });
   };
@@ -69,6 +73,6 @@ export const useMatchActions = () => {
   return {
     setViewed,
     addMatchAccount,
-    addMatchedAccount
+    addMatchedAccount,
   };
 };


TasoOneAsia:
Will review today

phone/src/apps/match/hooks/useMatchNotifications.ts
```diff
@@ -0,0 +1,42 @@
...
+    const notification = {
+      app: 'MATCH',
+      id: NOTIFICATION_ID,
+      sound: true,
+      title: t('MATCH.FEEDBACK.NEW_LIKE_FOUND'),
+      onClick: () => history.push(`/match/matches`),
+      content: name,
+      icon,
+      notificationIcon,
+    };
```
TasoOneAsia, line 35: We generally almost always create a callback reference using `useCallback` to prevent any unnecessary re renders in consuming components for this type of hook.

TasoOneAsia, line 35: Maybe remove this pending my completion of Notification API, @itschip thoughts?

resources/client/cl_match.ts
```diff
@@ -7,3 +8,11 @@ RegisterNuiProxy(MatchEvents.GET_MATCHES);
...
 RegisterNuiProxy(MatchEvents.CREATE_MY_PROFILE);
 RegisterNuiProxy(MatchEvents.UPDATE_MY_PROFILE);
+
+onNet(MatchEvents.SAVE_LIKES_BROADCAST, (result: CreateMatchBroadcast) => {
+    sendMatchEvent(MatchEvents.SAVE_LIKES_BROADCAST, result);
+});
+
+onNet(MatchEvents.CREATE_MATCH_ACCOUNT_BROADCAST, (result: FormattedProfile) => {
+    sendMatchEvent(MatchEvents.CREATE_MATCH_ACCOUNT_BROADCAST, result);
+});
```
TasoOneAsia, line 18: What's the reasoning behind needing a match account broadcast event? I believe we are doing active data fetching from NUI to receive pending matches on Match open. If I recall correctly, this shouldn't need to be pushed from the server.

resources/server/match/match.service.ts
```diff
@@ -42,21 +43,29 @@ class _MatchService {
...
-      await this.matchDB.saveLikes(identifier, reqObj.data);
+      await this.matchDB.saveLikes(player.getIdentifier(), reqObj.data);
     } catch (e) {
       matchLogger.error(`Failed to save likes, ${e.message}`);
       resp({ status: 'error', errorMsg: 'GENERIC_DB_ERROR' });
     }
 
     try {
-      const newMatches = await this.matchDB.findNewMatches(identifier, reqObj.data);
+      const newMatches = await this.matchDB.findNewMatches(player.getIdentifier(), reqObj.data);
```
TasoOneAsia, line 57: Please refactor to call player.getIdentifier() once.

```ts
const playerIdent = player.getIdentifier()
```

phone/src/apps/match/hooks/useMatchNotifications.ts
```diff
@@ -0,0 +1,42 @@
...
+  history.listen((location) => {
+    if (
+      matchPath(location.pathname, {
+        path: `/match/matches`,
+        exact: true,
+      })
+    ) {
+      removeId(NOTIFICATION_ID);
+    }
+  });
```
TasoOneAsia, line 23: React hooks are a form of functional programming. By not wrapping this in a `useEffect` this would create an assignment with every rerender of a consumer.

TasoOneAsia (commented):
Couple comments

resources/client/cl_match.ts
Patilhas: Exactly, when the match app is open. But if you open the app and then another persons account is created, that persons account is not gonna appear to you. This is like twitter, when a new tweet is created that tweet is send to the other player.

TasoOneAsia: Yes but for what reason do we need to push new accounts, this should be all handled on the server side?

phone/src/apps/match/hooks/useMatchNotifications.ts
itschip: Really depends if we want to do a new release before new notifications are done, and I think we should wait until that's done.

resources/server/match/match.service.ts
Patilhas: Fixed in next commit.

resources/client/cl_match.ts
Patilhas: Any idea other than that? Because, I only see giving the players the new account. Only if we use emitOnOpen so that it only requests when the app is open.

TasoOneAsia: emitOnOpen was primarily used in the past to signal to game scripts fetching needs to occur. Since then, we have taken an NUI primary approach where data fetching is generally handled through the proxy system and is started from CEF.

phone/src/apps/match/hooks/useMatchNotifications.ts
Patilhas: Any dependency?

resources/client/cl_match.ts
Patilhas: So do you have another approach that makes the same effect?

TasoOneAsia: The personal approach I would suggest is offloading matching in some server side handler. Then, whenever a client needs to request matches. It can do a simple fetch to a promise event on the server to get all matches when needed.

Patilhas: Ok, but when is that event is going to be triggered? How is the NUI going to figure out that a new possible match exists without going to check first?

TasoOneAsia: > Ok, but when is that event is going to be triggered? How is the NUI going to figure out that a new possible match exists without going to check first?

I'll need to review the match application in more depth, but from an architecture perspective we are attempting to reduce fire-and-forget game events events and are looking to replicate web behavior somewhat for data fetching.

phone/src/apps/match/hooks/useMatchService.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const useMatchService = () => {
+  const { setNotification } = useMatchNotifications();
+  const { addMatchAccount, addMatchedAccount } = useMatchActions();
+
+  const { state: profileLoading, contents: profileContent } = useRecoilValueLoadable(
+    matchState.myProfile,
+  );
+
+  const handleMatchBroadcast = ({ name }) => {
```
itschip, line 27: Do we really want to send an notification when they're in the app?

Patilhas: Fixed

Patilhas:
Forgot to call match service but it's fixed now

TasoOneAsia:
Will try and get around to this today.

TasoOneAsia:
@Patilhas Yea, looks like this won't fit into today's stuff. Will try and look tomorrow. In the meantime, CI is failing because of the following warnings. You should take a look at those warnings and try to resolve them. For reference,

pic

Patilhas:
> @Patilhas Yea, looks like this won't fit into today's stuff. Will try and look tomorrow. In the meantime, CI is failing because of the following warnings. You should take a look at those warnings and try to resolve them. For reference,
> 
> pic

I saw that warnings and I think that should ignore them but I'm not sure in useEffect of history is the right choice. Opinions?

TasoOneAsia:
These warnings are triggered by `eslint-react-hooks` which will help manage dependencies for you. Generally, it is almost always right.

Patilhas:
> These warnings are triggered by `eslint-react-hooks` which will help manage dependencies for you. Generally, it is almost always right.

In the case of useCallback I have to ignore because I want only on first mount. But useEffect I will have to see better.

phone/src/apps/match/hooks/useMatchNotifications.ts
```diff
@@ -38,9 +37,10 @@ export const useMatchNotifications = () => {
       notificationIcon,
     };
 
-    addNotificationAlert(notification)
+    addNotificationAlert(notification);
     addNotification(notification);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
itschip, line 42: Please don't ignore eslint warnings. Let's try and fix it instead.

Here you should add both `addNotificationAlert` and `addNotification` to the dependency array.

itschip (changes requested):

phone/src/apps/match/hooks/useMatchNotifications.ts
Patilhas: You think that I should also add icon and notificationIcon?

itschip: Yeah :)

itschip: Yeah

phone/src/apps/match/hooks/useMatchActions.ts
```diff
@@ -47,7 +51,24 @@ export const useMatchActions = () => {
...
+    if (matchesLoading) return;
+
+    set(matchState.profiles, (curVal) => [profile, ...curVal]);
+  });
+
+  const addMatchedAccount = async () => {
+    fetchNui<ServerPromiseResp<FormattedMatch[]>>(MatchEvents.GET_MATCHES).then((resp) => {
+      setMatches(resp.data);
+    });
+  };
```
itschip, line 67: Should add some error handling here.

Patilhas: Added status if is ok.

itschip (approved):

PR #605, opened 2022-02-13
Feat/rewrite messages

itschip:
Rewriting messages, *again*, sigh.

This time with group chats, and some improvments.

- [x] Relational data
- [x] Group chat
- [x] Notifications
- [x] Restoring conversations
- [x] UI improvments

Please feel free to test this branch out.

Here's some ideas, and concepts.

The proposal for storing each `user_identifier` in separate row, comes from Avarians PR proposal https://github.com/project-error/npwd/pull/531/commits/8ad1089cb0acfe55a130e3307e2c5e6186fd7017

I see now that this should solve a lot of weird issues, and reduce complexity, by A LOT (sorry Avarian). We’ve seen issues with the `unreadCount` being sent to the wrong client when fetching conversation. This is due to when we `reduce` conversations, we get the wrong `participant_identifier` unread count. Said change will fix this, even without reducing anything 😎

After a brainstorm with Avarian we have come up with a few ideas. We can now disregard the table above (I’ll still keep it). Also, groups are back!

We’ll create a relation between `message_conversations` and `message_participants`

An example for a `message_conversation` would be: 1 | 911+311 | 255

Where the first column, is the `conversation_id` , then `conversation_list` which contains all the phone numbers, and at last the `last_message_id`

The `message_participants` would just contain a row of a id, conversation_id and their phone number.

### General stuff, I guess? Or maybe ‘Functionality’

Most of the existing functionality, as sending messages, fetching messages, will most likely stay the same. Although there are cases where for example, transactions would be preferred, which I’ll have to look into closer. One example is when we’re creating conversations.

### API

People, or mostly roleplay fanatics wants a way to both listen to and send messages from different resources. The `onMessage` export was introduced recently, which went straight to hell. No surprise there. So let’s figure out how to make it work.

**onMessage**

The tricky export.

Let’s start with a code example, and I’ll try to explain the mess in my brain.

Concept: 

```lua
-- Let's say you have some service number for '911'
exports.npwd:onMessage("911", function(ctx)
	TriggerEvent("notifyPolice", ctx.message)

	ctx.respond("The police have been notified")
end)
```

The ctx contains the `source` from the player who sent the message. As well as the PreDBMessage data. This already exists in the export today.

...TBA

**emitMessage**

Concept:

```lua
-- Is this weird Lua formatting?
exports.npwd:emitMessage({ senderNumber = "7654321", targetNumber = "1234567", message = "Hello Taso" })
```

**createConversation**

This is pretty straight forward. Would take in two numbers, and npwd does some magic.

### DIFF
--- a/import.sql
+++ b/import.sql
@@ -160,21 +160,6 @@ CREATE TABLE `npwd_messages_participants`
     CONSTRAINT `message_participants_npwd_messages_conversations_id_fk` FOREIGN KEY (`conversation_id`) REFERENCES `npwd_messages_conversations` (`id`) ON UPDATE RESTRICT ON DELETE RESTRICT
 );
 
-CREATE TABLE IF NOT EXISTS `npwd_messages_conversations`
-(
-    `id`                     int(11)      NOT NULL AUTO_INCREMENT,
-    `user_identifier`        varchar(48)  NOT NULL,
-    `createdAt`              timestamp    NOT NULL DEFAULT CURRENT_TIMESTAMP(),
-    `updatedAt`              timestamp    NOT NULL DEFAULT CURRENT_TIMESTAMP() ON UPDATE CURRENT_TIMESTAMP,
-    `conversation_id`        varchar(512) NOT NULL,
-    `participant_identifier` varchar(48)  NOT NULL,
-    `label`                  varchar(60)           DEFAULT '',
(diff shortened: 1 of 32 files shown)


RuisSoftware:
Big boss! I will test this tonight to see if this solves my issue https://github.com/project-error/npwd/issues/599 or https://github.com/project-error/npwd/issues/600

PR #623, opened 2022-02-18
feat(twitter): bad words filter

Mojito-Fivem:
**Pull Request Description**

Adds a quick regex filter for blacklisted words on the twitter app

**Pull Request Checklist**:
* [x] Have you followed the guidelines in our Contributing document and Code of Conduct?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/project-error/npwd/pulls) for the same update/change?
* [x] Have you built and tested NPWD in-game after the relevant change?

PR #625, opened 2022-02-19
Feat/word filter

Mojito-Fivem:
**Pull Request Description**

Building on #623 , this adds a more generic way of filtering words which can be used anywhere if needed. Bad words are replaced with asterisks and a useWordFilter hook can be used to clean inputs of banned words.

**Pull Request Checklist**:

### DIFF
--- a/config.json
+++ b/config.json
@@ -40,6 +40,10 @@
       "wikipedia.org"
     ]
   },
+  "profanityFilter": {
+    "enabled": true,
+    "badWords": ["sus", "amogus"]
+  },
   "twitter": {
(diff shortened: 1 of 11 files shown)


itschip:
LGTM, but shouldn't we add this to messages and marketplace too?

TasoOneAsia:
Make sure to also PR an update to docs repo when able.

Mojito-Fivem:
Would it be worth adding some form of strict mode config which would just block messages from being sent as in the original PR

itschip (approved):

PR #665, opened 2022-03-17
Bug Fixes

AvarianKnight:
**Pull Request Description**

Fixes receiving messages for the wrong conversation

**Pull Request Checklist**:


AvarianKnight:
Some issues I'm going to look into tomorrow:

Deleting someone from contacts results in the phone crashing.

If Mojito doesn't fix the word filter putting the phone into an infinite crash loop I'll also look into that.

Mojito-Fivem:
I fixed the errors with the word filter handling but now there's still a crash loop here from the original error.

phone/src/apps/messages/hooks/useMessageActions.ts
```diff
@@ -89,6 +94,8 @@ export const useMessageActions = (): MessageActionProps => {
     (messageDto: Message) => {
       if (messageLoading !== 'hasValue') return;
 
+      if (conversationId !== messageDto.conversation_id) return;
```
itschip, line 97: We already handle this in [useMessageService](https://github.com/project-error/npwd/blob/develop/phone/src/apps/messages/hooks/useMessageService.ts#L29)

AvarianKnight: That's only used for sending messages, while this is for receiving messages.

PR #671, opened 2022-03-20
fix(messages): fix having the wrong target number on messages

AvarianKnight:
**Pull Request Description**
Fixes adding/calling someone, somewhat hacky but its the only place where the number gets sent

**Pull Request Checklist**:

### DIFF
--- a/phone/src/apps/messages/components/modal/MessageModal.tsx
+++ b/phone/src/apps/messages/components/modal/MessageModal.tsx
@@ -26,6 +26,7 @@ import { Call } from '@mui/icons-material';
 import { useMessageActions } from '../../hooks/useMessageActions';
 import GroupDetailsModal from './GroupDetailsModal';
 import Backdrop from '@ui/components/Backdrop';
+import { useMyPhoneNumber } from '@os/simcard/hooks/useMyPhoneNumber';
 
 const LARGE_HEADER_CHARS = 30;
 const MAX_HEADER_CHARS = 80;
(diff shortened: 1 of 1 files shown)


AvarianKnight:
PLEASE make sure to scrutinize this code heavily (mainly the WAS_REJECTED stuff) as I was between hundreds of thoughts on how to fix specific bugs and wouldn't be surprised if I left any code that could possibly regress

phone/src/apps/messages/hooks/state.ts
```diff
@@ -40,13 +42,18 @@ export const messageState = {
...
+      const messageConversations: MessageConversation[] = get(messageState.messageCoversations);      
       if (!searchValue) return messageConversations; // added this
 
-      const regExp = new RegExp(searchValue, 'gi');
+      const contacts: Contact[] = get(contactsState.contacts)
 
-      return messageConversations.filter((conversation) => conversation.participant.match(regExp));
+      return messageConversations.filter((messageConversation) => {
+        for (const contact of contacts) {
+          if (contact.display.includes(searchValue) && messageConversation.conversationList.includes(contact.number)) return true
```
AvarianKnight, line 52: Display could possibly be null

resources/server/messages/messages.db.ts
```diff
@@ -102,6 +103,11 @@ export class _MessagesDB {
 
     const result = <ResultSetHeader>results;
 
+    const updateConversation = `UPDATE npwd_messages_conversations SET updatedAt = current_timestamp() WHERE id = ?`;
+
+    // We don't await as this isn't important to the return data
+    DbInterface._rawExec(updateConversation, [dto.conversationId]);
```
jfrader, line 109: Maybe its better to call a side effect function and await there just to not have an unhandled rejection, then you log it somewhere and keep moving from here.

itschip (approved):

PR #693, opened 2022-03-28
tweak(net): global rate limiter for promise based net events

AvarianKnight:
Only really useful for Twitter search, but prevents users from being able to spam search


resources/server/lib/PromiseNetEvents/onNetPromise.ts
```diff
@@ -2,10 +2,18 @@ import { getSource } from '../../utils/miscUtils';
...
 import { ServerPromiseResp } from '../../../../typings/common';
+import { RateLimiter } from '../RateLimiter';
+import { GlobalRateLimiter } from '../GlobalRateLimiter';
 
 const netEventLogger = mainLogger.child({ module: 'events' });
 
-export function onNetPromise<T = any, P = any>(eventName: string, cb: CBSignature<T, P>): void {
+const globalRateLimiter = new GlobalRateLimiter(250);
+
+export function onNetPromise<T = any, P = any>(eventName: string, cb: CBSignature<T, P>, rateLimiter: RateLimiter = null): void {
```
antonstjernquist, line 12: Perhaps a better solution would to ONLY use the global rate limiter. 

And instead of passing a rateLimiter to the onNetPromise, we would pass an options object?

``` Javascript
{
  rateLimit: 250
}
```

itschip (approved):

Issue #711, opened 2022-04-06
Enhancement: Send location through messages


itschip:
### Application

Messages

### Dependency

No

### Feature Description

New context menu option to send your current coords through messages. Coords should not update over time.

AvaN0x:
Hello, I have just been looking at the code for embeds (mainly for the shared locations) and I was wondering something:
What would think about displaying the message content next to the shared location?

I'm asking this because from what I see, the [MessageEmbed](https://github.com/project-error/npwd/blob/481110f3710212f3dbe8e5672ac4163bf8c41d1b/phone/src/apps/messages/components/ui/MessageEmbed.tsx#L23) (and so [LocationEmbed ](https://github.com/project-error/npwd/blob/481110f3710212f3dbe8e5672ac4163bf8c41d1b/phone/src/apps/messages/components/ui/MessageEmbed.tsx#L63)) only has informations about the embed part of the message:
https://github.com/project-error/npwd/blob/481110f3710212f3dbe8e5672ac4163bf8c41d1b/phone/src/apps/messages/components/ui/MessageEmbed.tsx#L23
and even though the string `MESSAGES.LOCATION_MESSAGE` is sent as message content:
https://github.com/project-error/npwd/blob/481110f3710212f3dbe8e5672ac4163bf8c41d1b/phone/src/apps/messages/components/modal/MessageContextMenu.tsx#L68
this string will never get used because the embed is made to display `MESSAGES.LOCATION_MESSAGE` (so you can edit the message content in the database and nothing will change):
https://github.com/project-error/npwd/blob/481110f3710212f3dbe8e5672ac4163bf8c41d1b/phone/src/apps/messages/components/ui/MessageEmbed.tsx#L75

This would allow to send custom messages with locations through `emitMessage`, and could make the "Share location" button use the currently inputed text as message content or default to `MESSAGES.LOCATION_MESSAGE`.

Example of what it could look like: 
![image](https://user-images.githubusercontent.com/27494805/162399440-3897d5f9-483b-4b94-95df-b78d19ed896d.png)
*Disclaimer: I edited this with the devtools to see what it could look like*

Mojito-Fivem:
> That's why, even though the string `MESSAGES.LOCATION_MESSAGE` is sent as message content:
> https://github.com/project-error/npwd/blob/481110f3710212f3dbe8e5672ac4163bf8c41d1b/phone/src/apps/messages/components/modal/MessageContextMenu.tsx#L68
> 
> 
[...]

PR #712, opened 2022-04-06
feat(messages): location pinging

Mojito-Fivem:
**Pull Request Description**
Implements a share location feature in the messages app ctx menu as outlined in #711 . This allows people to share their location with the ability to set a waypoint. 

**Pull Request Checklist**:


Mojito-Fivem:
> Maybe instead of getting the player name for display, get contact name or phone number?

I disagree to be honest, the display name makes the most sense here

Patilhas:
> > Maybe instead of getting the player name for display, get contact name or phone number?
> 
> I disagree to be honest, the display name makes the most sense here, also for groups etc the contact name won't be the same for everyone. From the server it's either the character name or phone number we can get and I would say the name makes more sense.

It's just because, I'm going for a roleplay perspective right now. I'm talking to someone I don't know, it does really make sense to know his real name on the server when I didn't knew it before? It is a thing that can be easily changed. 

But you could also get the contact based on the number that sent the location on rendering if anyone wants to change it.

Mojito-Fivem:
From a roleplay perspective, If I have somebody's phone number I probably also know their name

Patilhas:
> From a roleplay perspective, If I have somebody's phone number I probably also know their name

There are a few exceptions especially in the crime world, but they exist. But as I said before, someone who has a little knowledge with the npwd codebase can change it easily.

Mojito-Fivem:
I'll leave it open to discussion and get some more opinions, as you said it shouldn't be too difficult to implement it that way.

AvaN0x:
> I'll leave it open to discussion and get some more opinions, as you said it shouldn't be too difficult to implement it that way.

I agree with @Patilhas, it can happen that you won't know the name of the person even if you have their number, or not their real name (in case of somebody who give a fake name for some reasons).
What I would suggest is either the conversation name, the phone number or a custom label that the user has to enter when sending coordinates *(could default to the phone number and still allow to set a custom label)*.

Mojito-Fivem:
Alright, it now uses the contact name or defaults to their phone number if you don't have them saved.

Mojito-Fivem:
image
image

PR #751, opened 2022-05-06
Feat/emit embed message

AvaN0x:
**Pull Request Description**

Edit embed messages to be able to display the message content:
image
image
The message content will only display if:
- `ContactEmbed`: the content is not the default one
- `LocationEmbed`: the content is not the default one, then it'll be displayed normally, else the default one will be displayed `Typography` (just like it did before)

---

Add the possibility to send `embed` with the export `emitMessage`. Exemple of use:
```lua
local sender = "123"
local target = "456-456-4568"
exports.npwd:emitMessage({
    senderNumber = sender,
    targetNumber = target,
    message = "Mission Row PD location",
    embed = {
        type = "location",
        coords = { 434.15, -981.71, 30.70 },
        phoneNumber = sender,
    }
})
```

**Pull Request Checklist**:

### DIFF
--- a/phone/src/apps/messages/components/ui/MessageEmbed.tsx
+++ b/phone/src/apps/messages/components/ui/MessageEmbed.tsx
@@ -23,29 +23,20 @@ type MessageEmbedType = {
 
 const MessageEmbed: React.FC<MessageEmbedProps> = ({ type, embed, isMine, message }) => {
   const embedType: MessageEmbedType = {
-    contact: <ContactEmbed embed={embed} isMine={isMine} message={message} />,
+    contact: <ContactEmbed embed={embed} isMine={isMine} />,
     location: <LocationEmbed embed={embed} isMine={isMine} message={message} />,
   };
 
   return <>{embedType[type]}</>;
 };
 
-const ContactEmbed = ({
-  isMine,
-  embed,
-  message,
-}: {
-  isMine: boolean;
-  embed: Contact;
-  message: string;
-}) => {
+const ContactEmbed = ({ isMine, embed }: { isMine: boolean; embed: Contact }) => {
   const [t] = useTranslation();
   const history = useHistory();
   const { pathname } = useLocation();
   const { getContactByNumber } = useContactActions();
 
   const showAddButton = !isMine && !getContactByNumber(embed?.number);
-  const showMessage = message !== t('MESSAGES.CONTACT_SHARED');
 
   const handleAddContact = () => {
     const referal = encodeURIComponent(pathname);
@@ -53,22 +44,19 @@ const ContactEmbed = ({
   };
 
   return (
-    <StyledMessage flexDirection="column">
-      {showMessage && <Box sx={{ width: '100%' }}>{message}</Box>}
-      <Box display="flex" alignItems="center" sx={{ width: '100%' }}>
-        <Box>
-          <Avatar src={embed?.avatar} />
-          <Typography>{embed?.display}</Typography>
-          <Typography>{embed?.number}</Typography>
-        </Box>
-        {showAddButton && (
-          <Box pl={'12px'}>
-            <Button fullWidth variant="contained" color="primary" onClick={handleAddContact}>
-              {t('GENERIC.ADD')}
-            </Button>
-          </Box>
-        )}
+    <StyledMessage>
+      <Box>
+        <Avatar src={embed?.avatar} />
+        <Typography>{embed?.display}</Typography>
+        <Typography>{embed?.number}</Typography>
       </Box>
+      {showAddButton && (
+        <Box pl={1}>
+          <Button fullWidth variant="contained" color="primary" onClick={handleAddContact}>
+            {t('GENERIC.ADD')}
+          </Button>
+        </Box>
+      )}
     </StyledMessage>
   );
 };
--- a/phone/src/apps/messages/components/ui/MessageEmbed.tsx
+++ b/phone/src/apps/messages/components/ui/MessageEmbed.tsx
@@ -89,16 +89,10 @@ const LocationEmbed = ({
     });
   };
 
-  const showMessageAsTypo = message === t('MESSAGES.LOCATION_MESSAGE');
-
   return (
     <StyledMessage>
       <Box>
-        {showMessageAsTypo ? (
-          <Typography>{message ?? t('MESSAGES.LOCATION_MESSAGE')}</Typography>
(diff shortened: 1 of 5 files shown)


resources/server/messages/middleware/emitMessage.ts
```diff
@@ -4,6 +4,11 @@ import MessagesService from '../messages.service';
...
 
 // FIXME: Fix this, oh wait, no one contributes anymore
-exp('emitMessage', async ({ senderNumber, targetNumber, message }: EmitMessageExportCtx) => {
-  await MessagesService.handleEmitMessage({ senderNumber, targetNumber, message });
+exp('emitMessage', async ({ senderNumber, targetNumber, message, embed }: EmitMessageExportCtx) => {
+  await MessagesService.handleEmitMessage({
+    senderNumber,
+    targetNumber,
+    message,
+    embed: embed ?? JSON.stringify(embed),
```
itschip, line 12: Would it not be more appropriate to just return an empty string if we don't have an embed? Instead of serializing "nothing".

itschip (changes requested):

resources/server/messages/middleware/emitMessage.ts
AvaN0x: My bad, I meant to use `&&` and not `??`.

phone/src/apps/messages/components/ui/MessageEmbed.tsx
```diff
@@ -14,53 +14,73 @@ interface MessageEmbedProps {
...
   return (
-    <StyledMessage>
-      <Box>
-        <Avatar src={embed?.avatar} />
-        <Typography>{embed?.display}</Typography>
-        <Typography>{embed?.number}</Typography>
-      </Box>
-      {showAddButton && (
+    <StyledMessage flexDirection="column">
+      {showMessage && <Box sx={{ width: '100%' }}>{message}</Box>}
```
Mojito-Fivem, line 57: Does this need a whole box with 100% width? 
image
You have a lot of empty space here

phone/src/apps/messages/components/ui/MessageEmbed.tsx
```diff
@@ -14,53 +14,73 @@ interface MessageEmbedProps {
...
-          <Button fullWidth variant="contained" color="primary" onClick={handleAddContact}>
-            {t('GENERIC.ADD')}
-          </Button>
+          <Avatar src={embed?.avatar} />
+          <Typography>{embed?.display}</Typography>
+          <Typography>{embed?.number}</Typography>
         </Box>
-      )}
+        {showAddButton && (
+          <Box pl={'12px'}>
```
Mojito-Fivem, line 65: Hardcoded pixel values, how well does this scale?

phone/src/apps/messages/components/ui/MessageEmbed.tsx
```diff
@@ -69,10 +89,16 @@ const LocationEmbed = ({ embed }: { embed: Location; isMine: boolean }) => {
     });
   };
 
+  const showMessageAsTypo = message === t('MESSAGES.LOCATION_MESSAGE');
+
   return (
     <StyledMessage>
       <Box>
-        <Typography>{t('MESSAGES.LOCATION_MESSAGE')}</Typography>
+        {showMessageAsTypo ? (
```
Mojito-Fivem, line 97: Shouldn't this always be typography

Mojito-Fivem (changes requested):

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: I tried to follow as most as I could the original design, the issue I had was with the `ADD` button position while having a message above.

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: From what I tested it fitted well. It is the same value as the X padding of the message bubble. I took it from devtools but I can edit it if you want.

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: I only used Typography when the message is the default one, so that when the message is custom, it'll show like the other messages do:
image
I can change it if you prefer

Mojito-Fivem: I think it should always match for the location sharing otherwise it doesn't quite look right

phone/src/apps/messages/components/ui/MessageEmbed.tsx
Mojito-Fivem: I'm not too sure about having messages show up in these embeds, it looks very out of place imo

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: Ok I'll edit it in next commit

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: Tbh I only added it in the ContactEmbed because I could, I don't really see the same use as for the LocationEmbed. I can remove if you prefer.

Mojito-Fivem: I think that would be for the better, in the location embeds I can see the case for it yeah

AvaN0x: Ok I'll remove it then

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: Changed in last commit

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: Changed in last commit

phone/src/apps/messages/components/ui/MessageEmbed.tsx
AvaN0x: Changed to `pl={1}` in last commit

itschip (approved):

PR #772, opened 2022-06-06
feat(messages): Remove Members from groups

SamShanks1:
**Pull Request Description**

User who created group can remove members from the group

When conversations are the phone number of the user who created it will be stored in the DB

The following SQL needs to be ran 
`ALTER TABLE npwd_messages_conversations ADD COLUMN `createdBy` varchar(48) NOT NULL DEFAULT `

**Pull Request Checklist**:
* [ ] Have you followed the guidelines in our Contributing document and Code of Conduct?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/project-error/npwd/pulls) for the same update/change?
* [ ] Have you built and tested NPWD in-game after the relevant change?

### DIFF
--- a/import.sql
+++ b/import.sql
@@ -156,6 +156,7 @@ CREATE TABLE `npwd_messages_conversations`
     `createdAt`         TIMESTAMP    NOT NULL DEFAULT current_timestamp(),
     `updatedAt`         TIMESTAMP    NOT NULL DEFAULT current_timestamp(),
     `last_message_id`   INT(11)      NULL     DEFAULT NULL,
+    `createdBy`         varchar(48)  NOT NULL DEFAULT ''
     `is_group_chat`     TINYINT(4)   NOT NULL DEFAULT '0',
     PRIMARY KEY (`id`) USING BTREE
 );
--- a/phone/src/apps/messages/components/modal/GroupDetailsModal.tsx
+++ b/phone/src/apps/messages/components/modal/GroupDetailsModal.tsx
@@ -3,6 +3,7 @@ import Modal from '@ui/components/Modal';
 import { Box, Button, Stack, Typography } from '@mui/material';
 import PersonIcon from '@mui/icons-material/Person';
 import PersonAddIcon from '@mui/icons-material/PersonAdd';
+import PersonRemoveIcon from '@mui/icons-material/PersonRemove';
 import { findParticipants } from '../../utils/helpers';
 import { useMyPhoneNumber } from '@os/simcard/hooks/useMyPhoneNumber';
 import { useContactActions } from '../../../contacts/hooks/useContactActions';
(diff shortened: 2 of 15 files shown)


SamShanks1:
One problem I just thought about will be groups that were created before this and don't have the "createdBy" defined, I could make it so any groups that don't have that will allow any member to remove another member. I also don't know if any errors will be created if "createdBy" is left blank

import.sql
```diff
@@ -156,6 +156,7 @@ CREATE TABLE `npwd_messages_conversations`
     `createdAt`         TIMESTAMP    NOT NULL DEFAULT current_timestamp(),
     `updatedAt`         TIMESTAMP    NOT NULL DEFAULT current_timestamp(),
     `last_message_id`   INT(11)      NULL     DEFAULT NULL,
+    `createdBy`         varchar(48)  NOT NULL DEFAULT ''
```
Mojito-Fivem, line 159: Missing a comma at the end of the line

resources/server/messages/messages.service.ts
```diff
@@ -285,6 +291,52 @@ class _MessagesService {
...
+      );
+      resp({ status: 'ok' });
+      const playerPhoneNumber = PlayerService.getPlayer(reqObj.source).getPhoneNumber();
+      const participants = reqObj.data.conversationList.split('+');
+      for (const participant of participants) {
+        if (reqObj.data.leaveGroup || participant !== playerPhoneNumber) {
+          const participantIdentifier = await PlayerService.getIdentifierByPhoneNumber(participant);
+          const participantPlayer = PlayerService.getPlayerFromIdentifier(participantIdentifier);
+          if (participantPlayer) {
+            if (!reqObj.data.leaveGroup && participant == reqObj.data.phoneNumber) {
```
Mojito-Fivem, line 312: `===` operator here?

Mojito-Fivem (changes requested):
Looks good so far, only issue I have is that members cannot re-join a group and a few minor things

TasoOneAsia:
Also appears CI is failing for the frontend due to a couple small ESLint complaints.

Screen Shot 2022-06-09 at 9 30 59 PM

phone/src/apps/messages/components/modal/GroupDetailsModal.tsx
```diff
@@ -51,15 +59,36 @@ const GroupDetailsModal: React.FC<GroupDetailsModalProps> = ({
...
+        sx={{
+          mt: 2,
+          display: 'flex',
+          justifyContent: 'space-evenly',
+        }}
+      >
+        <Button size="medium" onClick={leaveGroup}>
+          Leave Group
+        </Button>
+        {/* {myPhoneNumber === createdBy && (
```
itschip, line 88: To be removed?

SamShanks1:
Just need to do some last testing but this should be finished, @Mojito-Fivem would you prefer I do a separate PR for adding members to groups or keep it here

Mojito-Fivem:
It's all under the same feature and I think it would create more issues having this without the adding members functionality

SamShanks1:
Added system messages 

https://i.gyazo.com/d6d5615747923afe7b5ec819c934a1b8.mp4

- [ ] Add Light Mode Support

AlamoRoleplayFiveM:
Will there not be any chance of group calling?

arsiisra:
i have an idea. the owner can close the chat and only the owner him self can send message, or create a bol for chat user that can send message to chat.

PR #828, opened 2022-08-16
feat: splash screen with React.Suspense

antonstjernquist:


phone/src/os/apps/utils/createExternalAppProvider.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { useMemo } from 'react';
+import { createTheme, ThemeProvider, StyledEngineProvider, CircularProgress } from '@mui/material';
+import { deepMergeObjects } from '@shared/deepMergeObjects';
+import { usePhoneTheme } from '@os/phone/hooks/usePhoneTheme';
+import { IApp } from '../config/apps';
+import styled from '@emotion/styled';
+
+const colorShade = (col: any, amt: number) => {
```
itschip, line 8: Maybe it would be wise to wrap this in a useMemo?

itschip (approved):

phone/src/os/apps/utils/createExternalAppProvider.tsx
antonstjernquist: it's outside the function, so it wouldn't matter

PR #849, opened 2022-08-25
fix(phone/keyboard): Swap onFocus attributes to onMouseUp to fix weapon firing

INeedKnockup:
**Pull Request Description**

This PR aims to fix the weapon firing bug when a user clicks to focus an element.  The onFocus attribute used for the input fields was changed to onMouseUp to prevent firing the toggleKeys function simultaneously when the user pressed their mouse click (since firing the function at the same time the click is pressed will cause the bug to re-occur given the way that SetNuiFocusKeepInput functions).

### DIFF
--- a/phone/src/apps/contacts/components/ContactsApp.tsx
+++ b/phone/src/apps/contacts/components/ContactsApp.tsx
@@ -9,7 +9,7 @@ import { ContactPage } from './views/ContactsPage';
 import { ContactsThemeProvider } from '../providers/ContactsThemeProvider';
 import PersonAddIcon from '@mui/icons-material/PersonAdd';
 import Fab from '@mui/material/Fab';
-import { useHistory, useLocation } from 'react-router';
+import { useHistory, useLocation } from 'react-router-dom';
 import { Theme } from '@mui/material/styles';
 import makeStyles from '@mui/styles/makeStyles';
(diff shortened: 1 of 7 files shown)


itschip:
@INeedKnockup Can you please rebase this branch? :)

phone/src/apps/contacts/components/ContactsApp.tsx
```diff
@@ -9,10 +9,11 @@ import { ContactPage } from './views/ContactsPage';
 import { ContactsThemeProvider } from '../providers/ContactsThemeProvider';
 import PersonAddIcon from '@mui/icons-material/PersonAdd';
 import Fab from '@mui/material/Fab';
-import { useHistory, useLocation } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router';
```
itschip, line 12: Why do we switch to react-router here?

INeedKnockup: To be honest I'm not sure how this change became part of this PR, as only my changes

itschip: Please change it back :)

INeedKnockup: This has been fixed!

PR #897, opened 2022-10-09
Fix home route loading external apps too

### LINKED ISSUE #896: Bug: External apps get reloaded everytime u go to the home route
### Development Issue

Unsure

### Phone Version

Latest - 1.7,0

### Issue Description

External apps get reloaded everytime u go to the home route.
You can clearly see this because recoil gets angry that we're creating duplicated atoms.

### Issue Reproduction

1. Setup an external app (I tried [advertisements](https://github.com/npwd-community/npwd_advertisements))
2. Open any other app (doesn't matter which one, you only need to change route from home)
3. Return to home
4. Open devtools
5. Recoil should warn that atoms are being duplicated

### Evidence

Recoil warning
image

MM1212:
**Pull Request Description**

Home app was loading external apps (so is [Phone.tsx](https://github.com/project-error/npwd/blob/master/phone/src/Phone.tsx#L75)) with `useExternalApps()`, making them reload twice initially and everytime you went to the default route `/`.
This pr fixes #896 by only using the external apps already cached in the atom, and only let [Phone.tsx](https://github.com/project-error/npwd/blob/master/phone/src/Phone.tsx#L75) be the one loading them

**Pull Request Checklist**:

### DIFF
--- a/apps/phone/config/webpack.production.js
+++ b/apps/phone/config/webpack.production.js
@@ -59,6 +59,14 @@ module.exports = () => ({
           singleton: true,
           requiredVersion: deps['react-dom'],
         },
+        'react-router-dom': {
+          singleton: true,
+          requiredVersion: deps['react-router-dom'],
+        },
+        'recoil': {
+          singleton: true,
+          requiredVersion: deps['recoil'],
+        }
       },
     }),
     new HtmlWebpackPlugin({
--- a/apps/phone/src/apps/home/components/Home.tsx
+++ b/apps/phone/src/apps/home/components/Home.tsx
@@ -3,11 +3,12 @@ import { AppWrapper } from '@ui/components';
 import { Box } from '@mui/material';
 import { GridMenu } from '@ui/components/GridMenu';
 import { useApps } from '@os/apps/hooks/useApps';
-import { useExternalApps } from '@common/hooks/useExternalApps';
+import { useRecoilValue } from 'recoil';
+import { phoneState } from '@os/phone/hooks/state';
 
 export const HomeApp: React.FC = () => {
   const { apps } = useApps();
-  const externalApps = useExternalApps();
+  const externalApps = useRecoilValue(phoneState.extApps).filter(Boolean);
   return (
     <AppWrapper>
       <Box component="div" mt={6} px={1}>


MM1212:
Coming back to this found out that state isnt persistent at all, it just "refetches" the default value, which happens to be an async selector.
Fixing the first bug made the second one more "obvious"(for me at least).
I'll attempt to find a way to make recoil inside external apps detect the main RecoilRoot as I think is the only way to fix this :(

itschip:
It might that we are doing something wrong with the ModuleFederation plugin. It might be something to do with the shared package. I don't think t it should be defined in the external app webpack config. I could be wrong, but that's my first guess.

MM1212:
What if react-router-dom's Route is placed inside the external app itself as a child of RecoilRoot? Then the app can have its own RecoilRoot without breaking its state when unmounted by the Route

itschip:
@MM1212 Again, I think these are things ModuleFederation plugin can fix by configuring it. You could try the same approach for router-router/react-router-dom as for recoil, which I explained above.

MM1212:
> @MM1212 Again, I think these are things ModuleFederation plugin can fix by configuring it. You could try the same approach for router-router/react-router-dom as for recoil, which I explained above.

Ye you're right, just tried manually adding recoil & react-router-dom to the MF's shared configuration for both sides & it fixed the problem

MM1212:
So, in order for this to actual work, every ext app must remove its RecoilRoot, add to the webpack config recoil & react-router-dom and update phone's webpack config too.
Is there a better way to *propagate* this fix?

itschip:
> Is there a better way to propagate this fix?

I think this is how we want it to work.

MM1212:
I think u misunderstood xD
Right now, npwd-app-template state management doesn't work at all (when injected into the phone) as it [doesn't have its own RecoilRoot](https://github.com/project-error/npwd-app-template/blob/main/src/App.tsx#L120)
Fixing the `useExternalApps()` on the homescreen & adding the right MF's configuration to the phone isn't enough (although it doesn't break the apps either).
Should I also pr the right MF config to npwd-app-template? And all other official apps already created should also be updated.
Will be fixing the conflicts shortly too.

PR #996, opened 2023-04-25
update(client/cl_controls.lua): optimize thread

mikigoalie:


AvarianKnight:
This seems like a lot of un-needed complexity for very little performance gain

mikigoalie:
Well.. I think having 500 ms thread loop at all times wasn't great solution. The resmon can go up to .10 with phone open and in idle it's .05 - .08. With this commit I shaked off some resmon off of it.

Also I'm not sure about complexity. Now you just create thread for blocked keys when its needed and the pause menu got a little optimized. If it's not worth in your opinion then close it ^^

AvarianKnight:
I'm not sure where you're getting these numbers.

NPWD runs at 0.02 when open (notably it spikes to 0.07 for a second) and 0.01 when closed

mikigoalie:
lBPd

Mind asking me where did you pull the resmons from ? :)

TasoOneAsia:
Are you able to confirm that this specific path is what is contributing directly to the heightened resource time?  

I agree that this path could be optimized (slightly?) by initializing the checks according to the display state, however, considering how little code is being run and the relatively long the interval for this loop is, I am dubious of the practicable benefit. 

Also in terms of impl details, what's the logic behind the `collectgarbage()` invocation? Also, I am one of those people who dissuade the use of the `goto` in pretty much all programming language that implement them ([relevant XKCD](https://xkcd.com/292/)), if you give this PR another go, please try to _not use it_

mikigoalie:
It could differ from machine to machine, but having collectgarbage after blocked keys improved performance for me a little.

Regarding goto; fair. However Avarian said it is not in his opinion worth enough for merging so I will not reopen merge.
I just thought the lua part deserved proper changes, even if not THAT significant for performance.

Thanks for your input though,

AvarianKnight:
> It could differ from machine to machine, but having collectgarbage after blocked keys improved performance for me a little. With collecting garbage it went back to idle resmon, without it, I was able to spam reopening the phone and resmon went up to .15 (it was reported on discord many months ago).

Wonder if it might be a good idea to use Lua 5.4 and switch to generational gc.

I would also be perfectly fine with some of these changes, there are some here that make no sense (like the pause menu state change).

I also don't see a point in changing them from "anonymous" functions and creating them inline, i.e

```lua
CreateThread(function()
    while controlsDisabled do
		DisableAllControlActions(1)		  
		DisableAllControlActions(2)
        EnableControlAction(0, 22, true)
        EnableControlAction(0, 30, true)
        EnableControlAction(0, 31, true)
        EnableControlAction(0, 32, true)
        EnableControlAction(0, 33, true)
        Wait(0)
    end
end)
```

mikigoalie:
I guess it comes down to personal preference, feel free to let me know if it makes any difference, but in my opinion its cleaner to create threads from functions the way I did.

Again, its minor difference for performance, but I would still like to point out that resmon in idle can go to .03 - .05 in  idle, when the phone is closed. Unfortunately I don't know js well enough to do proper changes and optimizations, but I would be happy if it was something that would be adressed in 2.0 or upcoming updates.

PR #1020, opened 2023-07-11
Fix: Contacts App Input boxes were not disabling key actions

BackSH00TER:
**Pull Request Description**

On the Contacts App, ContactInfo component, the three `<Input />` fields were not disabling keys. 
If a player used WASD it would allow them to walk even while the input had focus, and other keys that were bound would also trigger.

Switched the component to use the `<InputBase />` component which is properly handling the keypresses. As a result of changing to this Input box, the component had to have the styles re-applied, so some style override changes were added.
BEFORE
image

AFTER
image

The Search box had the same issue so I used the `<SearchField />` component that other apps (like Messaging) are using.
BEFORE
image

AFTER
image

Recording showing that it no longer tries to make you move:
https://clipchamp.com/watch/JnjnPZK90pY

**Pull Request Checklist**:

### DIFF
--- a/apps/phone/src/apps/contacts/components/List/SearchContacts.tsx
+++ b/apps/phone/src/apps/contacts/components/List/SearchContacts.tsx
@@ -2,9 +2,10 @@ import React, { useEffect, useState } from 'react';
 
 import { useTranslation } from 'react-i18next';
 
+import { Box } from '@mui/material';
+import { SearchField } from '@ui/components/SearchField';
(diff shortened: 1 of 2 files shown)


BackSH00TER:
This also fixes the issues identified in: https://github.com/project-error/npwd/issues/1016

itschip (approved):

itschip (approved):

PR #1029, opened 2023-07-20
Add support for give contact details

BackSH00TER:
**Pull Request Description**

This PR adds support for giving contact details to other nearby players. 
It will get the closest player and will open their phone onto the fill new contact screen with the name and number prefilled. The user then just has to save it.

To use this, users would be expected to update their radial menu to use the `'npwd:GiveContactDetails` event.
Example w/ qb-radialmenu
```lua
{
    id = 'givenum',
    title = 'Give Contact Details',
    icon = 'address-book',
    type = 'client',
    event = 'npwd:GiveContactDetails',
    shouldClose = true
},
```

If one already exists it will show the already populated contact card and the user can update it from there.

Example:
https://clipchamp.com/watch/9zspIfzTHbG
- Sharing when nobody is around will notify you that nobody is nearby
- Sharing to a new player will open their phone to the contacts page with the info filled out
- Sharing to a player that already has your contacts opens to the page to edit the contact details

**Pull Request Checklist**:

### DIFF
--- a/apps/game/client/cl_contacts.ts
+++ b/apps/game/client/cl_contacts.ts
@@ -1,8 +1,72 @@
 import { ContactEvents } from '@typings/contact';
 import { RegisterNuiProxy } from './cl_utils';
 
+const QBCore = global.exports['qb-core'].GetCoreObject();
+const npwdExports = global.exports['npwd'];
+
+type Vector = {
+    x: number;
+    y: number;
+    z: number;
+  };
+
 RegisterNuiProxy(ContactEvents.PAY_CONTACT);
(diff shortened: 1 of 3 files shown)


apps/game/client/cl_contacts.ts
```diff
@@ -1,8 +1,72 @@
 import { ContactEvents } from '@typings/contact';
 import { RegisterNuiProxy } from './cl_utils';
 
+const QBCore = global.exports['qb-core'].GetCoreObject();
```
itschip, line 4: As this is supposed to be as standalone as possible, we cannot use any third party library. We have these exports for external use only.

itschip (changes requested):

PR #1125, opened 2024-01-19
Create new_da.json

iplayer1337fivem:
this is a new danish, since there is a da.json but its not in the phone. Hope it will get back, so we can get some fun with this as well. TY in advance

**Pull Request Description**

**Pull Request Checklist**:
* [ x] Have you followed the guidelines in our Contributing document and Code of Conduct?
* [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/project-error/npwd/pulls) for the same update/change?
* [ no] Have you built and tested NPWD in-game after the relevant change? Because I cant get it there. Im sorry. but I checked with the en.json and a compare tool, and no errors

### DIFF
--- a/apps/phone/src/locale/new_da.json
+++ b/apps/phone/src/locale/new_da.json
@@ -1,0 +1,360 @@
+{
+  "translation": {
+    "MISC": {
+      "TOP_LEVEL_ERR_TITLE": "Fatal fejl",
+      "TOP_LEVEL_ERR_ACTION": "Genindlæs",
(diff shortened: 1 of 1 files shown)


itschip:
Please update `da.json` instead :)

itschip:
You also need to add the language in the config

https://github.com/project-error/npwd/blob/master/apps/phone/src/config/default.json#L78

iplayer1337fivem:
fair enough, I had to do another way than usually, im sorry for wasting you time. :) I think I did it correct now. TY in advance

jfrader:
Not sure if you use this but its helpful to know what is missing, can probably try adding 'da' here
https://github.com/project-error/npwd/blob/master/apps/phone/i18n.missingKeys.js

